PR #1, opened 2016-03-30
MINIFI-2 Created a Util class to transform prospective config.yml into…

JPercivall:
… flow.xml and nifi.properties

### DIFF
--- a/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
+++ b/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
@@ -18,43 +18,34 @@
 package org.apache.nifi.minifi.bootstrap.util;
 
 
-import org.apache.nifi.connectable.Connection;
-import org.apache.nifi.connectable.StandardConnection;
-import org.apache.nifi.controller.FlowController;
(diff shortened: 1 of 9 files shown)


minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
```diff
@@ -0,0 +1,331 @@
...
+            writer.println("nifi.security.truststorePasswd=" + getValueString(securityProperties, TRUSTSTORE_PASSWORD_KEY));
+        } finally {
+            if (writer != null){
+                writer.flush();
+                writer.close();
+            }
+        }
+    }
+
+    private static void writeFlowXml(Map<String, Object> topLevelYaml, String path) throws IOException {
```
apiri: This should probably make use of the built in tools for marshalling/unmarshalling XML that NiFi uses.  With things like the reporting task for provenance and possibly more complex flows, it would give a good bit more flexibility moving forward.

JPercivall: Yup I definitely agree

JPercivall:
Updating and moving to MINIFI-15

PR #3, opened 2016-03-31
MINIFI-10 init commit of minifi-assembly

JPercivall:
I created a minifi-assembly that compiles and creates the initial LICENSE, NOTICE, bin and conf directories in a distribution. When deps get added to the bootstrap I believe it will need to be added to minifi-assembly.

I re-used much of the same scripts, pom.xml and dependencies.xml from NIFI (reconfigured for minifi) but I assume that since MiNiFi is a sub project of NiFi we don't need to provide a separate notice for work taken directly from it. I did add NiFi in the NOTICE for the assembly.


minifi-assembly/LICENSE
```diff
@@ -0,0 +1,1152 @@
...
+	OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+	NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+	HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+	WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+	FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+	OTHER DEALINGS IN THE SOFTWARE.
+
+
+
+This product bundles 'json2.js' which is available in the 'public domain'.
```
apiri: We will need to figure out which of these are or are not included.  Most of these should likely be removed and then added as they arise.

apiri:
@JPercivall need to clean up the LICENSE file to remove any unused items currently listed, otherwise looks good.

minifi-assembly/LICENSE
JPercivall: Yup you're right, I actually don't think any are used with just this PR. The only dependencies are internal modules.

apiri:
Looks good.  Will merge in from my review branch.  Thanks!

PR #4, opened 2016-03-31
MINIFI-9 initial commit for boostrapping/init process

JPercivall:
An initial commit for the minifi-boostrap. This is a repurposed nifi-bootstrap. It properly brings in the correct libs to start up but of course will relies on a MiNiFi.java class to start.


apiri:
Awesome, looks good here.  +1.  Can merge in off my local review branch.

PR #6, opened 2016-04-08
MINIFI-13 created a provenance reporting task to send provenance info…

JPercivall:
…rmation via S2S

### DIFF
--- a/minifi-nar-bundles/minifi-provenance-reporting-bundle/minifi-provenance-reporting-task/src/main/java/org/apache/nifi/minifi/provenance/reporting/ProvenanceReportingTask.java
+++ b/minifi-nar-bundles/minifi-provenance-reporting-bundle/minifi-provenance-reporting-task/src/main/java/org/apache/nifi/minifi/provenance/reporting/ProvenanceReportingTask.java
@@ -1,0 +1,457 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.nifi.minifi.provenance.reporting;
+
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TimeZone;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+
+import javax.json.Json;
+import javax.json.JsonArray;
+import javax.json.JsonArrayBuilder;
+import javax.json.JsonBuilderFactory;
+import javax.json.JsonObject;
+import javax.json.JsonObjectBuilder;
+import javax.net.ssl.SSLContext;
+
+import org.apache.nifi.annotation.behavior.Stateful;
+import org.apache.nifi.annotation.documentation.CapabilityDescription;
+import org.apache.nifi.annotation.documentation.Tags;
+import org.apache.nifi.annotation.lifecycle.OnScheduled;
+import org.apache.nifi.annotation.lifecycle.OnStopped;
+import org.apache.nifi.components.PropertyDescriptor;
+import org.apache.nifi.components.ValidationContext;
+import org.apache.nifi.components.ValidationResult;
+import org.apache.nifi.components.Validator;
+import org.apache.nifi.components.state.Scope;
+import org.apache.nifi.components.state.StateManager;
+import org.apache.nifi.components.state.StateMap;
+import org.apache.nifi.controller.ConfigurationContext;
+import org.apache.nifi.controller.status.PortStatus;
+import org.apache.nifi.controller.status.ProcessGroupStatus;
+import org.apache.nifi.controller.status.ProcessorStatus;
+import org.apache.nifi.controller.status.RemoteProcessGroupStatus;
+import org.apache.nifi.events.EventReporter;
+import org.apache.nifi.processor.exception.ProcessException;
+import org.apache.nifi.processor.util.StandardValidators;
+import org.apache.nifi.provenance.ProvenanceEventRecord;
+import org.apache.nifi.remote.Transaction;
+import org.apache.nifi.remote.TransferDirection;
+import org.apache.nifi.remote.client.SiteToSiteClient;
+import org.apache.nifi.reporting.AbstractReportingTask;
+import org.apache.nifi.reporting.ReportingContext;
+import org.apache.nifi.reporting.Severity;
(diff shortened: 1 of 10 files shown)


apiri:
reviewing

minifi-nar-bundles/minifi-provenance-reporting-bundle/minifi-provenance-reporting-task/src/main/test/java/org/apache/nifi/minifi/provenance/reporting/TestProvenanceReportingTask.java
```diff
@@ -0,0 +1,218 @@
...
+
+        assertEquals(1, dataSent.size());
+        final String msg = new String(dataSent.get(0), StandardCharsets.UTF_8);
+        System.out.println(msg);
+    }
+
+    public static FlowFile createFlowFile(final long id, final long fileSize, final Map<String, String> attributes) {
+        final Map<String, String> attrCopy = new HashMap<>(attributes);
+
+        return new FlowFile() {
```
apiri: Can likely use MockFlowFile instead of a new implementation.  Doesn't appear as though filesize is being used and the mock will allow setting of both the attributes and id.

minifi-nar-bundles/minifi-provenance-reporting-bundle/minifi-provenance-reporting-task/src/main/test/java/org/apache/nifi/minifi/provenance/reporting/TestProvenanceReportingTask.java
```diff
@@ -0,0 +1,218 @@
...
+        Mockito.when(initContext.getIdentifier()).thenReturn(UUID.randomUUID().toString());
+        Mockito.when(initContext.getLogger()).thenReturn(logger);
+
+
+        task.initialize(initContext);
+        task.onTrigger(context);
+
+        assertEquals(1, dataSent.size());
+        final String msg = new String(dataSent.get(0), StandardCharsets.UTF_8);
+        System.out.println(msg);
```
apiri: Either remove or introduce a logger for this instead of system out so we can control these items with the appropriate logging configs.

minifi-nar-bundles/minifi-provenance-reporting-bundle/minifi-provenance-reporting-task/src/main/java/org/apache/nifi/minifi/provenance/reporting/ProvenanceReportingTask.java
```diff
@@ -0,0 +1,443 @@
...
+            final byte[] data = jsonArray.toString().getBytes(StandardCharsets.UTF_8);
+            transaction.send(data, attributes);
+            transaction.confirm();
+            transaction.complete();
+
+            final long transferMillis = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
+            getLogger().info("Successfully sent {} Provenance Events to destination in {} ms; Transaction ID = {}; First Event ID = {}",
+                new Object[] {events.size(), transferMillis, transactionId, events.get(0).getEventId()});
+        } catch (final IOException e) {
+            getLogger().error("Failed to send Provenance Events to destination due to {}", new Object[] {e}, e);
```
apiri: I think some handling of the exception is needed, otherwise if there is an exception indices would likely be off with the following bit of code in terms of taking care of items even if nothing is sent.  In context of below, I think we should opt for at least once semantics.

minifi-nar-bundles/minifi-provenance-reporting-bundle/minifi-provenance-reporting-task/src/main/java/org/apache/nifi/minifi/provenance/reporting/ProvenanceReportingTask.java
```diff
@@ -0,0 +1,443 @@
...
+
+        if (firstEventId < 0) {
+            Map<String, String> state;
+            try {
+                state = context.getStateManager().getState(Scope.LOCAL).toMap();
+            } catch (IOException e) {
+                getLogger().error("Failed to get state at start up due to {}", e);
+                return;
+            }
+            if (state.containsKey(LAST_EVENT_ID_KEY)) {
```
apiri: Have to take into consideration how we provide a consistent view of provenance data in terms of provenance implementations such as the Volatile item which resets to 0 on a restart

minifi-nar-bundles/minifi-provenance-reporting-bundle/minifi-provenance-reporting-task/src/main/java/org/apache/nifi/minifi/provenance/reporting/ProvenanceReportingTask.java
```diff
@@ -0,0 +1,443 @@
...
+        final ProvenanceEventRecord lastEvent = events.get(events.size() - 1);
+        final String lastEventId = String.valueOf(lastEvent.getEventId());
+        try {
+            StateManager stateManager = context.getStateManager();
+            StateMap stateMap = stateManager.getState(Scope.LOCAL);
+            Map<String, String> newMapOfState = new HashMap<>();
+            newMapOfState.put(LAST_EVENT_ID_KEY, lastEventId);
+            stateManager.replace(stateMap, newMapOfState, Scope.LOCAL);
+        } catch (final IOException ioe) {
+            getLogger().error("Failed to update state to {} due to {}; this could result in events being re-sent after a restart of NiFi",
```
apiri: should be minifi

apiri:
Overall looks good, some minor changes.  Big concern is how we should handle the case of state and the possibility of a given producer.  Given that we are offloading to another system, I would be more fine with duplicating items that may not have rolled off than potentially missing them.

minifi-nar-bundles/minifi-provenance-reporting-bundle/minifi-provenance-reporting-task/src/main/java/org/apache/nifi/minifi/provenance/reporting/ProvenanceReportingTask.java
JPercivall: Ah good catch, I think the best we can do is to check whether the current max id in provenance is less than the last id queried (stored in state).

minifi-nar-bundles/minifi-provenance-reporting-bundle/minifi-provenance-reporting-task/src/main/java/org/apache/nifi/minifi/provenance/reporting/ProvenanceReportingTask.java
JPercivall: Yup I agree that it should be at least once, will change it to throw a ProcessException

minifi-nar-bundles/minifi-provenance-reporting-bundle/minifi-provenance-reporting-task/src/main/test/java/org/apache/nifi/minifi/provenance/reporting/TestProvenanceReportingTask.java
JPercivall: I agree, I am going change it to verify the message is getting transferred properly.

JPercivall:
Addressed the comments in the latest commit

minifi-nar-bundles/minifi-provenance-reporting-bundle/minifi-provenance-reporting-task/src/main/java/org/apache/nifi/minifi/provenance/reporting/ProvenanceReportingTask.java
```diff
@@ -251,6 +254,17 @@ public void onTrigger(final ReportingContext context) {
             if (state.containsKey(LAST_EVENT_ID_KEY)) {
                 firstEventId = Long.parseLong(state.get(LAST_EVENT_ID_KEY)) + 1;
             }
+
+            if(currMaxId < firstEventId){
+                getLogger().debug("Current provenance max id is {} which is less than what was stored in state as the last queried event, which was {}. This means the provenance restarted it's " +
```
apiri: super minor: it's -> its

apiri:
Looks good here, saw one minor grammatical thing, but otherwise a +1 from me.

PR #9, opened 2016-04-14
MINIFI-12 initial commit of http config change notifier

JPercivall:


JPercivall:
I decided against adding a utility library for this notifier because it is just a simple http server that allows users to post to it. There are many utilities to do that and creating a wrapper around one doesn't make a whole lot of sense when there isn't anything special/specific about the server.

minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/pom.xml
```diff
@@ -40,6 +40,24 @@ limitations under the License.
             <groupId>org.apache.nifi.minifi</groupId>
             <artifactId>minifi-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>ro.pippo</groupId>
+            <artifactId>pippo-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>ro.pippo</groupId>
+            <artifactId>pippo-tomcat</artifactId>
```
apiri: Was there a reason that Tomcat was chosen over some of the other alternatives?  No strong objection to Tomcat, but know that Jetty is used elsewhere.

pippo-core looks clear, but pippo-tomcat looks like it will need the LICENSE/NOTICE files updated as Tomcat seems to bring in a fair number of items

minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/configuration/RestChangeNotifier.java
```diff
@@ -0,0 +1,181 @@
...
+        webServerSettings.keystoreFile(properties.getProperty(KEYSTORE_LOCATION_KEY));
+        webServerSettings.keystorePassword(properties.getProperty(KEYSTORE_PASSWORD_KEY));
+        ((TomcatSettings) webServerSettings).keyAlias(properties.getProperty(KEY_ALIAS_KEY));
+
+        String tomcatHome = properties.getProperty(TOMCAT_HOME_KEY);
+        if(tomcatHome != null) {
+            ((TomcatSettings) webServerSettings).baseFolder(tomcatHome);
+        }
+
+        pippo.getApplication().getPippoSettings().overrideSetting("tomcat.baseFolder","./target/tomcat_home/");
```
apiri: base folder needs some default.  seems this was remnant of development

minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/configuration/RestChangeNotifier.java
```diff
@@ -0,0 +1,181 @@
...
+
+
+    @Override
+    public Set<ConfigurationChangeListener> getChangeListeners() {
+        return configurationChangeListeners;
+    }
+
+    @Override
+    public boolean registerListener(ConfigurationChangeListener listener) {
+        configurationChangeListeners.add(listener);
```
apiri: likely want to return the value from the add.  in the case of the Set it will indicate if a no entry was added or if it was already stored in the Set.

minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/pom.xml
JPercivall: It was part due to not knowing enough about the differences between Tomcat and Jetty, as well as having a bit of trouble configuring Jetty to do https. If Jetty is much better suited to our needs I will work through the problems I was having initially.

minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/configuration/RestChangeNotifier.java
JPercivall: Yup definitely a remnant of dev

minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/src/main/java/org/apache/nifi/minifi/configuration/RestChangeNotifier.java
JPercivall: Ah duh, thanks for catching that

minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/pom.xml
apiri: Ah gotcha. In the interim as we are kind of prototyping things out let's not go through solving time sinks for "optimization" at this juncture. Maybe make another ticket to address at a later time? Implementation specifics are hidden behind the interface so should be okay to swap.

apiri:
This code will need to be relocated to the bootstrap module.  I had things heading down the wrong path with my initial commit of file change notifier and remedied it with the work for MINIFI-14/PR #10

minifi-nar-bundles/minifi-framework-bundle/minifi-framework/minifi-runtime/pom.xml
JPercivall: That works for me, the only change that would be impact the user would be the changing of a couple config properties (like a removal of the base folder)

JPercivall:
I will adjust the module and address the comments once MINIFI-14 is merged in

apiri:
@JPercivall This was taken care of with #13, correct?  If so, could you close this one out?

JPercivall:
Yes it was, will close, Thanks @apiri

PR #15, opened 2016-04-21
MINIFI-17 Adding error handling of configurations that fail to start …

JPercivall:
…and a couple other small changes

### DIFF
--- a/minifi-assembly/pom.xml
+++ b/minifi-assembly/pom.xml
@@ -261,7 +261,7 @@ limitations under the License.
         <!-- nifi.properties: web properties -->
         <nifi.web.war.directory>./lib</nifi.web.war.directory>
         <nifi.web.http.host />
-        <nifi.web.http.port>8080</nifi.web.http.port>
+        <nifi.web.http.port>8081</nifi.web.http.port>
         <nifi.web.https.host />
         <nifi.web.https.port />
         <nifi.jetty.work.dir>./work/jetty</nifi.jetty.work.dir>
--- a/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
+++ b/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
@@ -57,6 +57,7 @@ import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import org.apache.nifi.minifi.bootstrap.configuration.ConfigurationChangeException;
 import org.apache.nifi.minifi.bootstrap.configuration.ConfigurationChangeListener;
 import org.apache.nifi.minifi.bootstrap.configuration.ConfigurationChangeNotifier;
 import org.apache.nifi.minifi.bootstrap.util.ConfigTransformer;
@@ -1051,6 +1052,15 @@ public class RunMiNiFi {
     @SuppressWarnings({"rawtypes", "unchecked"})
     public void start() throws IOException, InterruptedException {
 
+        final String confDir = getBootstrapProperties().getProperty(CONF_DIR_KEY);
+        final File configFile = new File(getBootstrapProperties().getProperty(MINIFI_CONFIG_FILE_KEY));
+        try {
+            performTransformation(new FileInputStream(configFile), confDir);
+        } catch (ConfigurationChangeException e) {
+            defaultLogger.error("The config file is malformed, unable to start.", e);
+            return;
+        }
+
         Tuple<ProcessBuilder, Process> tuple = startMiNiFi();
         if (tuple == null) {
             cmdLogger.info("Start method returned null, ending start command.");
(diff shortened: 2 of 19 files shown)


apiri:
reviewing

minifi-assembly/pom.xml
```diff
@@ -261,7 +261,7 @@ limitations under the License.
         <!-- nifi.properties: web properties -->
         <nifi.web.war.directory>./lib</nifi.web.war.directory>
         <nifi.web.http.host />
-        <nifi.web.http.port>8080</nifi.web.http.port>
```
apiri: Should be converted back to the typical 8080 default.

JPercivall: I switched it because we're planning on getting rid of the UI anyway so currently it is primarily for debugging purposes and it's easier to debug when it doesn't have a conflicting default port as NiFi. If it makes more sense to switch it back I can then I would also change it in the ConfigTransformer (defaults to 8081 during transformation).

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
```diff
@@ -1080,15 +1111,31 @@ public void start() throws IOException, InterruptedException {
...
                     final boolean previouslyStarted = getNifiStarted();
                     if (!previouslyStarted) {
-                        defaultLogger.info("MiNiFi never started. Will not restart MiNiFi");
-                        return;
+                        final File swapConfigFile = getSwapFile(defaultLogger);
+                        if(swapConfigFile.exists()){
+                            defaultLogger.info("Swap file exists, MiNiFi failed trying to change configuration. Reverting to old configuration.");
+
+                            final String confDir = getBootstrapProperties().getProperty(CONF_DIR_KEY);
+                            ((MiNiFiConfigurationChangeListener) changeListener).performTransformation(new FileInputStream(swapConfigFile), confDir);
```
apiri: Would prefer to just declare the field of this class to just be a MiNiFiConfigurationChangeListener instead of having to do the cast.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
```diff
@@ -1379,7 +1430,7 @@ private void saveFile(final InputStream configInputStream, File configFile) {
             }
         }
 
-        private void performTransformation(InputStream configIs, String configDestinationPath) {
+        public void performTransformation(InputStream configIs, String configDestinationPath) {
```
apiri: Understand the need for increased visibility but think protected or package private would be sufficient.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
JPercivall: I agree, changing

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
JPercivall: That is true, I will make it package private

apiri:
Overall looks good with the comments mentioned on the commit.  I did run into some problems with malformed config.yml files though that resulted in NPEs in the ConfigTransformer functionality.  In this case, a yml of:

``` yaml
Flow Controller:
    name: MiNiFi Flow
    comment:
```

Resulted in an NPE and caused the notifying process to have issues.  A stacktrace of the File based (although the REST based would have similar issues) is below:

```
2016-04-22 10:58:00,210 ERROR [File Change Notifier Thread] o.apache.nifi.minifi.bootstrap.RunMiNiFi Unable to successfully transform the provided configuration
java.lang.NullPointerException: null
        at org.apache.nifi.minifi.bootstrap.util.ConfigTransformer.writeNiFiProperties(ConfigTransformer.java:175) ~[minifi-bootstrap-0.0.1-SNAPSHOT.jar:0.0.1-SNAPSHOT]
        at org.apache.nifi.minifi.bootstrap.util.ConfigTransformer.transformConfigFile(ConfigTransformer.java:155) ~[minifi-bootstrap-0.0.1-SNAPSHOT.jar:0.0.1-SNAPSHOT]
        at org.apache.nifi.minifi.bootstrap.RunMiNiFi$MiNiFiConfigurationChangeListener.performTransformation(RunMiNiFi.java:1435) [minifi-bootstrap-0.0.1-SNAPSHOT.jar:0.0.1-SNAPSHOT]
        at org.apache.nifi.minifi.bootstrap.RunMiNiFi$MiNiFiConfigurationChangeListener.handleChange(RunMiNiFi.java:1408) [minifi-bootstrap-0.0.1-SNAPSHOT.jar:0.0.1-SNAPSHOT]
        at org.apache.nifi.minifi.bootstrap.configuration.FileChangeNotifier.notifyListeners(FileChangeNotifier.java:70) [minifi-bootstrap-0.0.1-SNAPSHOT.jar:0.0.1-SNAPSHOT]
        at org.apache.nifi.minifi.bootstrap.configuration.FileChangeNotifier.run(FileChangeNotifier.java:125) [minifi-bootstrap-0.0.1-SNAPSHOT.jar:0.0.1-SNAPSHOT]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [na:1.8.0_77]
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) [na:1.8.0_77]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) [na:1.8.0_77]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) [na:1.8.0_77]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [na:1.8.0_77]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [na:1.8.0_77]
        at java.lang.Thread.run(Thread.java:745) [na:1.8.0_77]
```

JPercivall:
Just added a fairly large commit more throughly handling configuration errors and other corner cases involving the swap and config files as well as the nifi.properties and flow.xml.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/configuration/ConfigurationChangeNotifier.java
```diff
@@ -52,6 +52,5 @@
     /**
      * Provide the mechanism by which listeners are notified
      */
-    void notifyListeners();
-
+    ListenerHandleResult[] notifyListeners();
```
apiri: Would prefer to use a Collection opposed to an array

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/configuration/notifiers/FileChangeNotifier.java
```diff
@@ -63,15 +71,23 @@
...
-            } catch (IOException ex) {
-                throw new IllegalStateException("Unable to read the changed file " + configFile, ex);
+                listenerHandleResults[count] = new ListenerHandleResult(listener);
+            } catch (IOException | ConfigurationChangeException ex) {
+                listenerHandleResults[count] = new ListenerHandleResult(listener, ex);
             }
+            logger.info("Listener notification result:" + listenerHandleResults[count].toString());
+            count ++;
         }
+        return new ListenerHandleResult[0];
```
apiri: this should be listenerHandleResults.  this is only returning a new empty array.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/configuration/notifiers/RestChangeNotifier.java
```diff
@@ -227,15 +235,27 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
                     }
                 }
                 setConfigFile(configBuilder.substring(0,configBuilder.length()-1));
-                notifyListeners();
-                writeOutput(response, POST_TEXT, 200);
+                ListenerHandleResult[] listenerHandleResults = notifyListeners();
+
+                writeOutput(response, getPostText(listenerHandleResults), 200);
```
apiri: With the new collection of results, it seems like the returned status code should reflect the status of the attempted notifications.

JPercivall: Good call, I changed it to return the text status of the listeners but didn't change the status code. I can't tell whether the listener failed due to client or server error so I will just return 200 if all succeeded but 500 if any failed.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/configuration/ConfigurationChangeNotifier.java
JPercivall: Yup I agree

JPercivall:
Pushed a commit changing the ConfigurationChangeNotifier to return an Collection instead of Array

apiri:
Checking out the latest commit and doing some testing.

apiri:
Looks good.  Tried a number of different bad scenarios and it seemed to work as anticipated.  +1 from me.  I'll leave the squashing/merging to you at your convenience.

JPercivall:
Awesome, thanks @apiri I will squash, merge and close now

PR #17, opened 2016-04-26
MINIFI-30 added changes to the config transformer to support multiple…

JPercivall:
… processors and exfiling using a processor

### DIFF
--- a/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
+++ b/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
@@ -20,6 +20,22 @@ package org.apache.nifi.minifi.bootstrap.util;
 
 import org.apache.nifi.controller.FlowSerializationException;
 import org.apache.nifi.minifi.bootstrap.configuration.ConfigurationChangeException;
+import org.apache.nifi.minifi.bootstrap.exception.InvalidConfigurationException;
+import org.apache.nifi.minifi.bootstrap.util.schema.ComponentStatusRepositorySchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.ConfigSchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.ConnectionSchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.ContentRepositorySchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.CorePropertiesSchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.FlowControllerSchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.FlowFileRepositorySchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.ProcessorSchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.ProvenanceReportingSchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.ProvenanceRepositorySchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.RemoteInputPortSchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.RemoteProcessingGroupSchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.SecurityPropertiesSchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.SensitivePropsSchema;
+import org.apache.nifi.minifi.bootstrap.util.schema.SwapSchema;
 import org.apache.nifi.stream.io.ByteArrayOutputStream;
 import org.w3c.dom.DOMException;
 import org.w3c.dom.Document;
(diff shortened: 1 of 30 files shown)


apiri:
starting review

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
```diff
@@ -219,6 +223,7 @@ private static void writeNiFiProperties(Map<String, Object> topLevelYaml, Output
             Map<String, Object> componentStatusRepo = (Map<String, Object>) topLevelYaml.get(COMPONENT_STATUS_REPO_KEY);
             Map<String, Object> securityProperties = (Map<String, Object>) topLevelYaml.get(SECURITY_PROPS_KEY);
             Map<String, Object> sensitiveProperties = (Map<String, Object>) securityProperties.get(SENSITIVE_PROPS_KEY);
+            Map<String, Object> provenanceProperties = (Map<String, Object>) securityProperties.get(PROVENANCE_REPORTING_KEY);
```
apiri: Believe this should be topLevelYaml and not security properties.

apiri: Upon further inspection I see this is unused and the provenance properties are gathered in createFlowXml.  Likely just remove unless you would prefer to gather these maps up front.

JPercivall: Will remove, I think I originally thought something in the nifi.properties needed it but I guess I was mistaken

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
```diff
@@ -428,13 +436,26 @@ private static void addProcessGroup(final Element parentElement, Map<String, Obj
...
+            if (processors != null) {
+                for (Object processorConfig : processors) {
+                    addProcessor(element, (Map<String, Object>) processorConfig);
+                }
+            }
 
-            Map<String, Object> remoteProcessingGroup = (Map<String, Object>) topLevelYaml.get(REMOTE_PROCESSING_GROUP_KEY);
-            addRemoteProcessGroup(element, remoteProcessingGroup);
+            Map<String, Object> exfilComponent = (Map<String, Object>) topLevelYaml.get(EXFIL_COMPONENT_KEY);
+            if ("Site-To-Site".equals(getValueString(exfilComponent, TYPE_KEY))) {
```
apiri: minor:  would be nice to make this a constant as it is used in two different places for checks

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
```diff
@@ -428,13 +436,26 @@ private static void addProcessGroup(final Element parentElement, Map<String, Obj
...
+                addRemoteProcessGroup(element, exfilComponent);
+            } else if ("PROCESSOR".equals(getValueString(exfilComponent, TYPE_KEY))) {
```
apiri: minor:  would be nice to make this a constant as it is used in a few different places for the same purpose

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
```diff
@@ -616,15 +630,21 @@ private static void addConnection(final Element parentElement, Map<String, Objec
             addTextElement(element, "labelIndex", "1");
             addTextElement(element, "zIndex", "0");
 
-            addTextElement(element, "sourceId", "Processor");
+            addTextElement(element, "sourceId", getValueString(connectionProperties, SOURCE_NAME_KEY));
```
apiri: For properties like these that are required, we should have an indication that it is needed and an associated exception if it is not present.

apiri:
Overall looks like a good start to accomplish what is needed for this effort.  Some additional validation is needed to enforce the required properties such that we can catch them before attempting to begin the flow anew after a transformation.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
apiri: One additional thought I had was to have another class that perhaps does the value checking and enforces the config YAML and then passes a value object (a collection of all the backing maps) to the ConfigTransformer.  Long term, we may not have this particular process and could isolate the actual validation and checking elsewhere.  In lieu of the manual checks, we may be able to make use of something like [Kwalify](http://www.kuwata-lab.com/kwalify/) (no clue as to its quality) which might help in cutting down on the onerous and manual if statements for each individual scalar.

JPercivall: That is a very good way forward and I will refactor this to be two classes. I'll look in to Kwalify and google a bit to see what others are using for YAML schema validation.

JPercivall:
@apiri just pushed out a new commit refactoring the validation to a separate classes. There are merge conflicts with master in ConfigTransformer that I will address tomorrow.

JPercivall:
@apiri rebased and pushed

JPercivall:
Thinking a bit more, I don't see a reason to not allow multiple RPGs either. Will make the adjustment

JPercivall:
@apiri added support for multiple RPGs and multiple Input Ports within those RPGs

apiri:
This looks good overall.  I would avoid the sheer ClassCast exception handling and prefer to do instanceof checks instead.  Reads a bit cleaner.

Also, for the majority of cases, we have a pattern that is roughly:
```Java
  value getKeyAsType(valueMap, key, target, wrapperName, reason, optional, defaultValue) {
        value = valueMap.get(key);
        if (value == null) {
            if (optional && defaultValue != null) {
                return defaultValue;
            }
        } else {
            targetClass = target.getClass()
            if (targetClass.isInstance(value)) {
                return value;
            } else {
                addValidationIssue(key, wrapperName, reason + targetClass);
            }
        }
        return null;
    }
```

This with a few overloaded signatures would allow cutting down on a significant number of lines and centralize logic throughout.  I saw a few special cases, which should likely be left to their own, but it looks like we can standardize the bulk of the validation.

JPercivall:
@apiri just pushed out a new commit centralizing the logic for getting the various values

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/common/BaseSchema.java
```diff
@@ -51,18 +55,80 @@ public void addValidationIssue(String keyName, String wrapperName, String reason
...
-    public void addIssuesIfNotNull(BaseSchema baseSchema) {
-        if (baseSchema != null) {
-            validationIssues.addAll(baseSchema.getValidationIssues());
+
+    public <T> T getMapAsType(Map valueMap, String key, Class targetClass, String wrapperName, boolean required) {
+        Object obj = valueMap.get(key);
+        return interpretValueAsType(obj, key, targetClass, wrapperName, required);
+    }
+
+    public void instantiateList(List list, String simpleListType, Class targetClass, String wrapperName){
```
apiri: Is it fair to say that this is more of a transform/mapping to type for the already existing list?

apiri:
Hey @JPercivall, changes look great and overall much cleaner.  Just had the question regarding naming/function of instantiateList, but otherwise looks good to merge in.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/common/BaseSchema.java
JPercivall: I agree

JPercivall:
@apiri added change to the method name and squashed the commits, if that still looks good I can merge it or you

apiri:
@JPercivall rock and roll 🎸  with that change.  Although I think something went awry with your imports. Travis had an issue as well, looks like the same locally.

> [INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /Users/apiri/Development/code/apache/nifi-minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/RemoteProcessingGroupSchema.java:[57,12] error: cannot find symbol
[ERROR]   symbol:   method instantiateList(List<RemoteInputPortSchema>,String,Class<RemoteInputPortSchema>,String)
  location: class RemoteProcessingGroupSchema
/Users/apiri/Development/code/apache/nifi-minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/ConfigSchema.java:[71,12] error: cannot find symbol
[ERROR]   symbol:   method instantiateList(List<ProcessorSchema>,String,Class<ProcessorSchema>,String)
  location: class ConfigSchema
/Users/apiri/Development/code/apache/nifi-minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/ConfigSchema.java:[76,12] error: cannot find symbol
[ERROR]   symbol:   method instantiateList(List<ConnectionSchema>,String,Class<ConnectionSchema>,String)
  location: class ConfigSchema
/Users/apiri/Development/code/apache/nifi-minifi/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/ConfigSchema.java:[81,12] error: cannot find symbol

apiri:
@JPercivall was waiting for Travis to do the build asynchronously but had a moment to pull down and build.  Looks good now.  Super minor, but there is a trailing quote on your commit message.  May want to scrap that before merging.

Thanks for tackling this!

PR #19, opened 2016-06-17
MINIFI-40 - Defaulting to empty string for keystore, truststore, sslP…

brosander:
…rotocol

### DIFF
--- a/minifi-bootstrap/pom.xml
+++ b/minifi-bootstrap/pom.xml
@@ -79,6 +79,13 @@ limitations under the License.
             <groupId>org.apache.nifi</groupId>
             <artifactId>nifi-framework-core</artifactId>
             <version>0.6.0</version>
+            <exclusions>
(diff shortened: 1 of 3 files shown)


minifi-bootstrap/pom.xml
```diff
@@ -43,6 +43,10 @@ limitations under the License.
         </dependency>
         <dependency>
             <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-properties</artifactId>
```
JPercivall: Instead of adding another pom dependency for one static helper method I would prefer to just create a new private one that does it:
    private static boolean nullOrEmpty(String input) {
        return input == null || input.isEmpty();
    }

JPercivall:
+1 

Fixes the issue, and tested by doing secure S2S transmission back to a core NiFi instance. Thanks for addressing this issue @brosander! I will merge it in.

PR #20, opened 2016-06-27
MINIFI-43 Fixing provenance instantiation and making some default con…

JPercivall:
…structors private to prevent same problem in other classes

I made the no parameter constructor private when there are required properties with no default value.

### DIFF
--- a/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/ConfigSchema.java
+++ b/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/ConfigSchema.java
@@ -52,11 +52,7 @@ public class ConfigSchema extends BaseSchema {
 
     private ProvenanceRepositorySchema provenanceRepositorySchema;
 
-    private ConfigSchema() {
-    }
-
     public ConfigSchema(Map map) {
-        Object obj ;
         flowControllerProperties = getMapAsType(map, FLOW_CONTROLLER_PROPS_KEY, FlowControllerSchema.class, TOP_LEVEL_NAME, true);
 
         coreProperties = getMapAsType(map, CORE_PROPS_KEY, CorePropertiesSchema.class, TOP_LEVEL_NAME, false);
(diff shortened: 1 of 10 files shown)


apiri:
Reviewing.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/ConfigSchema.java
```diff
@@ -52,7 +52,7 @@
 
     private ProvenanceRepositorySchema provenanceRepositorySchema;
 
-    public ConfigSchema() {
+    private ConfigSchema() {
```
apiri: Changing to private, should this c'tor just be removed?

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/ConnectionSchema.java
```diff
@@ -44,7 +44,7 @@
     private String flowfileExpiration = "0 sec";
     private String queuePrioritizerClass = "";
 
-    public ConnectionSchema() {
+    private ConnectionSchema() {
```
apiri: Same as above

apiri:
Hey @JPercivall.

Changes look good to resolve the listed issue in the ticket.  Aside from that, I think we can probably clean up the private constructors as it appears we have a defined constructor for each using the map.

Also, would you be able to cleanup https://github.com/JPercivall/nifi-minifi/blob/1c5c2d49e5e0477a0b6dbb58cbd2c3ecf73b4a52/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/ConfigSchema.java#L59?  I realize it wasn't added by this commit, but might be nice to remove it while making changes.

JPercivall:
@apiri I agree with your comments. I was looking for the easy fix (just make them private) but didn't think it all the way through. Thanks

I made the adjustment and removed the extra object in ConfigSchema

apiri:
@JPercivall looks good.  Rock and roll :guitar: +1

JPercivall:
Thanks @apiri I merged it into master

PR #21, opened 2016-06-27
MINIFI-41 - CLI utility for template.xml -> YAML and YAML validation

brosander:


JPercivall:
The CLI utility should not be in the bootstrap module. The bootstrap module is specifically for things that get run as part of the bootstrap process. This CLI utility is unlike any other portion of the project and should be moved into it's module.

In order for it to access the Schema pojo's it may be best to bring those out into another module as well.

brosander:
@JPercivall we wouldn't need to move things around that much if we put lib/bootstrap/* on the template cli's classpath, is that an acceptable approach?

JPercivall:
@brosander its not just the functionality we need to take into account, it's putting the classes in logic places for further development. If the schema pojos are no longer just used in the bootstrap then it doesn't make sense to leave them in there (and make any module that wants to use them depend on the entire bootstrap module). 

Also, since this is a CLI utility that is not dependent on an instance being deployed I don't think it makes sense to bundle this in a distribution. When thinking specifically in terms of Design and Deploy, the template will be created and transformed (using this CLI) in one location and then distributed to many MiNiFi instances. 

@apiri may want to weigh in as well.

brosander:
@JPercivall I moved out all of the schema logic to minifi-commons/minifi-commons-schema.  Then I pulled out the config main and supporting classes into minifi-toolkit/minifi-toolkit-configuration.  Finally I created an assembly in minifi-toolkit/minifi-toolkit-assembly.

minifi-commons/minifi-commons-schema/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+            <version>1.17</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-framework-core</artifactId>
+            <version>0.6.0</version>
+            <exclusions>
+                <!-- excluding for IDE purposes, this jar doesn't wind up in the lib/bootstrap folder so isn't safe to use in bootstrap project -->
```
JPercivall: I don't think this is applicable for the new assembly.

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/SchemaLoader.java
```diff
@@ -0,0 +1,52 @@
...
+                throw new SchemaLoaderException("Provided YAML configuration is not a Map");
+            }
+        } catch (YAMLException e ) {
+            throw new IOException(e);
+        } finally {
+            sourceStream.close();
+        }
+    }
+
+    public static ConfigSchema loadConfigSchema(InputStream sourceStream) throws IOException, SchemaLoaderException {
```
JPercivall: Since this is just a wrapper around loadYamlAsMap and has the requirement that the input is Yaml, I'd prefer to have "Yaml" in somewhere in the method name.

JPercivall:
In the schema module, all of those classes extending BaseSchema are in the same package as the utility classes (like SchemaLoader, SchemaSaver, etc.) and it makes it a bit hard to digest. I would prefer them to be separate.

minifi-toolkit/minifi-toolkit-assembly/NOTICE
```diff
@@ -0,0 +1,21 @@
...
+The following binary components are provided under the Apache Software License v2
+
+  (ASLv2) Apache NiFi
+    The following NOTICE information applies:
+      Apache NiFi
+      Copyright 2014-2016 The Apache Software Foundation
+
+  (ASLv2) Jetty
+    The following NOTICE information applies:
+       Jetty Web Container
```
JPercivall: Is Jetty packaged in the toolkit-assembly?

minifi-commons/minifi-commons-schema/pom.xml
brosander: Good point, removed

minifi-toolkit/minifi-toolkit-assembly/NOTICE
brosander: Nope, removing

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/BaseSchema.java
```diff
@@ -15,13 +15,17 @@
...
 import java.lang.reflect.InvocationTargetException;
+import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 public class BaseSchema {
```
JPercivall: I think I originally should have made this class Abstract. One of the main things is the toMap method. In this class all it does it return a LinkedHashMap and that just feels off to me. I see that others are overriding it and just using this to get a LinkedHashMap. Also I think it is counter intuitive to have the ability to initialize a BaseSchema. 

Overall I believe all the "Value Access/Interpretation helper methods" should be make static and toMap would be abstract. Thoughts?

brosander: So I agree that the toMap method is a little light atm, the one thing I liked about defining it in the superclass was to centralize the decision to use a LinkedHashMap (to preserve insertion order and allow YAML to have attributes in the desired order).

I figured that it was possible we would want to add common attributes at the top level at some point in which case that would be a good place for it.

In terms of the helper methods, I don't have a problem with making them static.

minifi-toolkit/minifi-toolkit-assembly/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>minifi-toolkit</artifactId>
+        <groupId>org.apache.nifi.minifi</groupId>
+        <version>0.0.1-SNAPSHOT</version>
+    </parent>
+    <artifactId>minifi-toolkit-assembly</artifactId>
+    <packaging>pom</packaging>
+    <description>This is the assembly of Apache NiFi - MiNiFi</description>
```
brosander: Updated description

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/BaseSchema.java
brosander: So those helper methods do access the internal state (validation issues).  I can tweak them to take that as an argument but it seems like leaving them as instance methods makes sense for now.

JPercivall: Right the "Validation Issue helper methods" access the validation issues, which is something every schema will have, so it's good to keep them implemented how they are. 

The "Value Access/Interpretation helper methods" are all methods that do not necessarily rely on the internal representation of each schema but instead are helper methods for accessing/interpreting values. So they can be static.

For toMap I agree that it's good to encourage the use of LinkedHashMap. Maybe a protected method getMap for the schemas to use. Then toMap would be abstract.

brosander: The getKeyAsType() and interpretValueAsType() both add validation issues and need the schema reference to do so.  Those are used by the rest of the Value Access/Interpretation helper methods which also need the reference.

JPercivall: Ah very true, ok I'm on board with not making them static.

brosander: Using map supplier to instantiate maps, toMap() is static now, as well as the helper methods that don't need the state of the schema to do their job.

minifi-toolkit/minifi-toolkit-assembly/src/main/resources/config.sh
```diff
@@ -0,0 +1,133 @@
...
+        fi
+    fi
+    # if command is env, attempt to add more to the classpath
+    if [ "$1" = "env" ]; then
+        [ "x${TOOLS_JAR}" =  "x" ] && [ -n "${JAVA_HOME}" ] && TOOLS_JAR=$(find -H "${JAVA_HOME}" -name "tools.jar")
+        [ "x${TOOLS_JAR}" =  "x" ] && [ -n "${JAVA_HOME}" ] && TOOLS_JAR=$(find -H "${JAVA_HOME}" -name "classes.jar")
+        if [ "x${TOOLS_JAR}" =  "x" ]; then
+             warn "Could not locate tools.jar or classes.jar. Please set manually to avail all command features."
+        fi
+    fi
```
JPercivall: Is this block left over from the minifi.sh? I tried running it and got this:
Joseph-Percivall:bin jpercivall$ ./config.sh env

Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_74.jdk/Contents/Home
MiNiFi Toolkit home: /Users/jpercivall/projects/edge/nifi-minifi/minifi-toolkit/minifi-toolkit-assembly/target/minifi-0.0.1-SNAPSHOT-bin/minifi-toolkit-0.0.1-SNAPSHOT


Usage:

java org.apache.nifi.minifi.toolkit.configuration.ConfigMain <command> options

Valid commands include:
transform: Transform template xml into MiNiFi config YAML
validate: Validate config YAML

minifi-toolkit/minifi-toolkit-configuration/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.nifi.minifi</groupId>
+            <artifactId>minifi-commons-schema</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi</groupId>
+            <artifactId>nifi-framework-core</artifactId>
+            <version>0.6.0</version>
```
JPercivall: This should be added to the Dependency Management portion of the root pom so that it shares the nifi version set there.

JPercivall:
Unless if I'm missing them, validation issues are not created in Schemas created by DTOs. I don't think we can assume the template will always contain valid values.

minifi-commons/minifi-commons-schema/pom.xml
JPercivall: (Accidentally commented on the wrong pom originally) This should probably be removed from this pom too since it will be used in both assemblies.

brosander:
You're right, my mistake

brosander:
@JPercivall I think I've addressed your feedback

brosander:
working on tests for validation now

minifi-toolkit/minifi-toolkit-configuration/src/test/resources/Splunk_Processors_with_Listen_TCP.xml
```diff
@@ -0,0 +1,18 @@
...
...lowableValues><displayName>error</displayName><value>error</value></allowableValues><defaultValue>info</defaultValue><description>The Log Level to use when logging the Attributes</description><displayName>Log Level</displayName><dynamic>false</dynamic><name>Log Level</name><required>true</required><sensitive>false</sensitive><supportsEl>false</supportsEl></value></entry><entry><key>Log Payload</key><value><allowableValues><displayName>true</displayName><value>true</value></allowableValues><allowableValues><displayName>false</displayName><value>false</value></allowableValues><defaultValue>false</defaultValue><description>If true, the FlowFile's payload will be logged, in addition to its attributes; otherwise, just the Attributes will be logged.</description><displayName>Log Payload</displayName><dynamic>false</dynamic><name>Log Payload</name><required>true</required><sensitive>false</sensitive><supportsEl>false</supportsEl></value></entry><entry><key>Attributes to Log</key><value><description>A comma-separated list of Attributes to Log. If not specified, all attributes will be logged.</description><displayName>Attributes to Log</displayName><dynamic>false</dynamic><name>Attributes t
```
JPercivall: We currently only bundle the NiFi standard processors by default (other nars can be added manually) but we should probably only refer to those to avoid confusion.

brosander: @JPercivall not seeing another pre-existing template with ssl service configured, I was trying to have a realistic example.  Think this can be addressed in a README or in the xml and yaml themselves?

apiri:
The packaging being separate certainly makes sense.  The utility is kind of similar in context as that of the "externals" in the core NiFi project.  Seems fair to keep it distinct given how instances would likely be established.  We can evaluate including it in the resulting assembly as we move forward.  May make sense to include it in these early stages to get people running.

minifi-toolkit/minifi-toolkit-configuration/src/test/resources/Splunk_Processors_with_Listen_TCP.xml
JPercivall: I'd rather have a non-realistic testing flow than one that gives the illusion that we bundle processors that we don't. I can create a template to use.

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
```diff
@@ -0,0 +1,342 @@
...
+
+    public int validate(String[] args) {
+        if (args.length != 2) {
+            printValidateUsage();
+            return ERR_INVALID_ARGS;
+        }
+        try (InputStream inputStream = pathInputStreamFactory.create(args[1])) {
+            try {
+                ConfigSchema configSchema = SchemaLoader.loadConfigSchemaFromYaml(inputStream);
+                if (!configSchema.isValid()) {
```
JPercivall: It would be nice to print a line saying the flow is valid

JPercivall:
Probably outside the scope of this PR/Ticket but the ConfigSchema should raise a validation issue when multiple components have the same name. It will be pretty common in templates to have multiple processors with the same name (ie. LogAttribute).

JPercivall:
@brosander did you get a template/config working that didn't include a queue prioritizer? When I run with '' for the queue prioritizer field in the config.yml I see the error below. I cross referenced with the flow xml as it appears in the NiFi instance I made the template in and I think there should be an if statement on line 541 of ConfigTransformer where it won't add the "queuePrioritizerClass" element at all if it's empty.

org.apache.nifi.controller.FlowSynchronizationException: java.lang.IllegalArgumentException: Unable to set prioritizer : java.lang.ClassNotFoundException: 
	at org.apache.nifi.controller.StandardFlowSynchronizer.sync(StandardFlowSynchronizer.java:317) ~[nifi-framework-core-0.6.0.jar:0.6.0]
	at org.apache.nifi.controller.FlowController.synchronize(FlowController.java:1287) ~[nifi-framework-core-0.6.0.jar:0.6.0]
	at org.apache.nifi.persistence.StandardXMLFlowConfigurationDAO.load(StandardXMLFlowConfigurationDAO.java:72) ~[nifi-framework-core-0.6.0.jar:0.6.0]
	at org.apache.nifi.controller.StandardFlowService.loadFromBytes(StandardFlowService.java:629) ~[nifi-framework-core-0.6.0.jar:0.6.0]
	at org.apache.nifi.controller.StandardFlowService.load(StandardFlowService.java:479) ~[nifi-framework-core-0.6.0.jar:0.6.0]
	at org.apache.nifi.web.contextlistener.ApplicationStartupContextListener.contextInitialized(ApplicationStartupContextListener.java:79) ~[na:na]
	at org.eclipse.jetty.server.handler.ContextHandler.callContextInitialized(ContextHandler.java:800) ~[jetty-server-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.servlet.ServletContextHandler.callContextInitialized(ServletContextHandler.java:444) ~[jetty-servlet-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.server.handler.ContextHandler.startContext(ContextHandler.java:791) ~[jetty-server-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.servlet.ServletContextHandler.startContext(ServletContextHandler.java:294) ~[jetty-servlet-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.webapp.WebAppContext.startWebapp(WebAppContext.java:1349) ~[jetty-webapp-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.webapp.WebAppContext.startContext(WebAppContext.java:1342) ~[jetty-webapp-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.server.handler.ContextHandler.doStart(ContextHandler.java:741) ~[jetty-server-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.webapp.WebAppContext.doStart(WebAppContext.java:505) ~[jetty-webapp-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68) [jetty-util-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.util.component.ContainerLifeCycle.start(ContainerLifeCycle.java:132) [jetty-util-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.util.component.ContainerLifeCycle.doStart(ContainerLifeCycle.java:114) [jetty-util-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.server.handler.AbstractHandler.doStart(AbstractHandler.java:61) [jetty-server-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68) [jetty-util-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.util.component.ContainerLifeCycle.start(ContainerLifeCycle.java:132) [jetty-util-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.server.Server.start(Server.java:387) [jetty-server-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.util.component.ContainerLifeCycle.doStart(ContainerLifeCycle.java:114) [jetty-util-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.server.handler.AbstractHandler.doStart(AbstractHandler.java:61) [jetty-server-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.server.Server.doStart(Server.java:354) [jetty-server-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68) [jetty-util-9.2.11.v20150529.jar:9.2.11.v20150529]
	at org.apache.nifi.web.server.JettyServer.start(JettyServer.java:662) [nifi-jetty-0.6.0.jar:0.6.0]
	at org.apache.nifi.minifi.MiNiFi.<init>(MiNiFi.java:138) [minifi-runtime-0.0.1-SNAPSHOT.jar:0.0.1-SNAPSHOT]
	at org.apache.nifi.minifi.MiNiFi.main(MiNiFi.java:233) [minifi-runtime-0.0.1-SNAPSHOT.jar:0.0.1-SNAPSHOT]
Caused by: java.lang.IllegalArgumentException: Unable to set prioritizer : java.lang.ClassNotFoundException: 
	at org.apache.nifi.controller.StandardFlowSynchronizer.addProcessGroup(StandardFlowSynchronizer.java:1009) ~[nifi-framework-core-0.6.0.jar:0.6.0]
	at org.apache.nifi.controller.StandardFlowSynchronizer.sync(StandardFlowSynchronizer.java:271) ~[nifi-framework-core-0.6.0.jar:0.6.0]
	... 27 common frames omitted

JPercivall:
@brosander were you able to get a flow running using a flow that was transformed from a template that uses an SSL context service? I'm looking at the config and code and I believe there are going to be problems. When the template is exported the properties that "Identify Controller service" have the value of the ID. This causes problems for us because we use Names as our IDs and the SSL context service will have the name "SSL-Context-Service". 

We would either need to scan through every property to determine if it was referring to the SSL context and then change it to the proper name; or implement a way to set the ID of the SSL context service. I dislike both options: option 1 would require knowledge/parsing of arbitrary processor properties, option 2 would be implementing ID just for this one purpose.

It may be better to hold back the commit adding SSL controller service support and create an improvement jira that you can use that commit towards later.

brosander:
@JPercivall the Queue Prioritizer field defaults to empty string if it's not set, meaning it shouldn't matter whether the field is in the yaml or not, it will still wind up being the same empty string when parsed (https://github.com/apache/nifi-minifi/blob/master/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/schema/ConnectionSchema.java#L56)

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
brosander: Added a print statement when schema is valid

brosander:
@JPercivall Removing the ssl context service code for now.

JPercivall:
Right it'll be an empty field but I think from the flow.xml side, there is a difference between "" and it not being there. When I ran it without a prioritizer I observed that error, can you run it properly as it currently is?

brosander:
I see what you're saying now, I was just surprised that that would be different than a hand-edited YAML with an empty or nonexistent queue prioritizer field.  I can change that but I don't think it's really within the scope of this jira.

JPercivall:
I don't think it is different than with hand-edited YAML and you're probably right about it being outside the scope of this Jira. Either it was just overlooked, I forgot that that was a limitation or I messed up thing when configuring it.

JPercivall:
For RemoteProcessGroups, you need to set the destination name of the connection to be the id of the port (not the name of it). Check out this config file (just rename to yaml, Github doesn't let me upload yaml files). It was generate using the toolkit script and it produces the error:
`
org.apache.nifi.controller.FlowSynchronizationException: java.lang.RuntimeException: Found Invalid Connectable ID for Destination: response
`
Once the destination name is properly set to the id of the port (b23a4621-cf19-42e6-967c-ffd3716e6a24) it properly runs.

[invalidDestintationName.txt](https://github.com/apache/nifi-minifi/files/346816/invalidDestintationName.txt)

minifi-toolkit/minifi-toolkit-assembly/NOTICE
```diff
@@ -0,0 +1,16 @@
...
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+===========================================
+Apache Software License v2
+===========================================
+
+The following binary components are provided under the Apache Software License v2
+
+  (ASLv2) Apache NiFi
```
JPercivall: Apache NiFi's notice includes more than just this. It should be this[1]:
> 
> Apache NiFi
> Copyright 2014-2016 The Apache Software Foundation
> 
> This product includes software developed at
> The Apache Software Foundation (http://www.apache.org/).
> 
> This product includes the following work from the Apache Hadoop project:
> 
> BoundedByteArrayOutputStream.java adapted to SoftLimitBoundedByteArrayOutputStream.java

[1] https://github.com/apache/nifi/blob/nifi-0.6.0/NOTICE

brosander: I took that from [the minifi assembly notice](https://github.com/apache/nifi-minifi/blob/a1d2fd3fe6b0b2cdd50394e2451aa10d5c9c81e9/minifi-assembly/NOTICE#L13).  Should I update that one too or address in separate jira?

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
JPercivall: Unless if I'm missing something, I see a print statement confirmation for "transform" but I don't see one for "validate".

brosander: You're right, addressing

minifi-bootstrap/src/test/resources/Working_with_Logs.yml
```diff
@@ -0,0 +1,253 @@
...
+    name: user logs
+    comment: ''
+    max concurrent tasks: 1
+    use compression: false
+  - id: 70f8fe07-ecff-4ed4-b050-5c8e4e419eec
+    name: app logs
+    comment: ''
+    max concurrent tasks: 1
+    use compression: false
+Provenance Reporting:
```
JPercivall: Nit pick: this yaml hasn't been updated to have the Provenance reporting section removed (this is currently invalid due to missing fields).

brosander: It wasn't a great example anyways, using the templates I ran through successfully instead.

JPercivall:
+1

Did a contrib-check build, reviewed code and all comments were addressed, tested functionality and it all looks good. Thanks for the contribution @brosander I will squash it and merge it in.

PR #23, opened 2016-07-03
MINIFI-38 Removing reliance on JettyServer in order to add a flow sta…

JPercivall:
…tus reporting end point. This also removes the UI and adds a 'flowstatus' option to minifi.sh to get information on the current flow from the terminal.


apiri:
reviewing

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
```diff
@@ -228,6 +233,9 @@ public static void main(String[] args) throws IOException, InterruptedException
             case "env":
                 runMiNiFi.env();
                 break;
+            case "flowstatus":
+                runMiNiFi.statusReport(args[1]);
```
apiri: May want to do a check on this and provide usage information if no args are provided

minifi-commons/minifi-utils/src/main/java/org/apache/nifi/minifi/commons/status/util/StatusReportPopulator.java
```diff
@@ -0,0 +1,375 @@
...
+import org.apache.nifi.minifi.commons.status.system.HeapStatus;
+import org.apache.nifi.minifi.commons.status.system.SystemDiagnosticsStatus;
+import org.apache.nifi.minifi.commons.status.system.SystemProcessorStats;
+
+import java.util.Collections;
+import java.util.Date;
+import java.util.LinkedList;
+import java.util.List;
+
+public class StatusReportPopulator {
```
apiri: Looks like this is used for creating canned objects.  Should this be in src/test?

minifi-commons/minifi-utils/src/main/test/java/org/apache/nifi/minifi/commons/status/TestStatusReport.java
```diff
@@ -0,0 +1,88 @@
...
+import static org.apache.nifi.minifi.commons.status.util.StatusReportPopulator.addConnectionStatus;
+import static org.apache.nifi.minifi.commons.status.util.StatusReportPopulator.addControllerServiceStatus;
+import static org.apache.nifi.minifi.commons.status.util.StatusReportPopulator.addExpectedRemoteProcessGroupStatus;
+import static org.apache.nifi.minifi.commons.status.util.StatusReportPopulator.addInstanceStatus;
+import static org.apache.nifi.minifi.commons.status.util.StatusReportPopulator.addProcessorStatus;
+import static org.apache.nifi.minifi.commons.status.util.StatusReportPopulator.addReportingTaskStatus;
+import static org.apache.nifi.minifi.commons.status.util.StatusReportPopulator.addSystemDiagnosticStatus;
+import static org.junit.Assert.assertEquals;
+
+public class TestStatusReport {
```
apiri: This appears to be in the wrong folder structure.  src/main/test/java instead of src/test/java

apiri:
Hey @JPercivall,

Quite the marathon PR, but certainly needed steps to get MiNiFi off on its own and those status objects, while tedious, will come in handy moving forward.  Had a few comments, but it all looks pretty good.  Testing out the flowStatus functionality looked good and I think this gives us a pretty nice foundation to build on.  We will need to adjust things around a bit in terms of dependencies and such, but the functionality of the original ticket looks good, so let's get this in and iterate.

Had a few minor comments, but overall looks good to go.

minifi-commons/minifi-utils/src/main/java/org/apache/nifi/minifi/commons/status/util/StatusReportPopulator.java
JPercivall: The problem is that it's used in tests for both minifi-utils (TestStatusReport) as well as framwork-core (TestStatusConfigReporter). If it's in src/test then it isn't available in the other module. I could just replicate it in the other module if you think that's better.

minifi-commons/minifi-utils/src/main/test/java/org/apache/nifi/minifi/commons/status/TestStatusReport.java
JPercivall: Yup you're right, will fix.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
JPercivall: Good call, will add

JPercivall:
@apiri I just pushed a commit addressing two of the comments. If you think it's best I can move the StatusReportPopulator into both modules, just let me know.

minifi-commons/minifi-utils/src/main/java/org/apache/nifi/minifi/commons/status/util/StatusReportPopulator.java
apiri: For future reference, you can specify a classifier of tests or type with a jar plugin to make use of those without having to include them in the source artifact.  Not worth the overhead for one class, but a nice bit to keep in your pocket should this need grow into many such dependencies.

apiri:
Seems good here.  Will merge.  Thanks for tackling this.

PR #24, opened 2016-07-05
MINIFI-49 Made TestRestChangeNotifier upload file system independent.

JPercivall:


minifi-bootstrap/src/test/java/org/apache/nifi/minifi/bootstrap/configuration/notifiers/util/TestRestChangeNotifierCommon.java
```diff
@@ -84,6 +78,14 @@ public void testFileUpload() throws Exception {
 
         assertEquals("The result of notifying listeners:\nMockChangeListener successfully handled the configuration change\n", response.body().string());
 
-        assertEquals(new String(Files.readAllBytes(file.toPath())), mockChangeListener.getConfFile());
+        // Ignoring the system dependent line ending.
+        String confFile = mockChangeListener.getConfFile();
+        if (confFile.endsWith("\n") || confFile.endsWith("\r")) {
```
apiri: Do you think it would be better to just use a StringUtils#trim on the returned conf file?

apiri:
Looks good overall, good catch on the OS dependency.  Just thought maybe a StringUtils#trim would be a bit cleaner and straightforward.

minifi-bootstrap/src/test/java/org/apache/nifi/minifi/bootstrap/configuration/notifiers/util/TestRestChangeNotifierCommon.java
JPercivall: You are totally right, will change

JPercivall:
@apiri, I pushed a new commit that uses trim instead of my unnecessary logic, lol, if you wanna give a final approval and merge.

apiri:
Ha, no worries.  The logic was all right just verbose. :wink:

Will do a final check/build and merge if that all looks good.

PR #25, opened 2016-07-05
MINIFI-22 Removing lingering parts of UI

JPercivall:
Most of this work was done with MINIFI-38 but there were still some lingering transitive dependencies being brought in that needed to be removed.

### DIFF
--- a/minifi-assembly/pom.xml
+++ b/minifi-assembly/pom.xml
@@ -121,6 +121,16 @@ limitations under the License.
             <artifactId>minifi-provenance-repository-nar</artifactId>
             <type>nar</type>
         </dependency>
+        <dependency>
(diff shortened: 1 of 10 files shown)


apiri:
reviewing

apiri:
Good finds and cleanup.  I additionally removed the following exclusion as it was being referenced in the build as not being triggered:  https://github.com/JPercivall/nifi-minifi/blob/d98d0970bf1929dcd3da13a51c7dd4b27d08680d/minifi-assembly/src/main/assembly/dependencies.xml#L38

Will push.  Thanks!

JPercivall:
Ah good catch with the exclusion. Thanks for the review @apiri

PR #26, opened 2016-07-05
MINIFI-48 - Omitting queuePrioritizerClass from flow if it isn't set …

brosander:
…in schema


JPercivall:
+1

Did a contrib-check build, looked over the code and verified functionality. Thanks for addressing this @brosander

PR #30, opened 2016-07-07
MINIFI-54 - Fixing duplicate validation for connections, remote proce…

brosander:
…ssing groups


JPercivall:
+1

Did contrib check build, verified the code and verified using the transform/validation util. Thanks for the fix @brosander

PR #32, opened 2016-09-07
MINIFI-104 - Making connection ids filesystem friendly, unique

brosander:


apiri:
Reviewing

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
```diff
@@ -546,6 +550,37 @@ protected static void addConnection(final Element parentElement, ConnectionSchem
...
+    /**
+     * Will replace all characters not in [A-Za-z0-9_] with _
+     *
+     * This has potential for collisions so it will also append numbers as necessary to prevent that
+     *
+     * @param ids id map of already incremented numbers
+     * @param name the name
+     * @return a unique filesystem-friendly id
+     */
+    protected static String getUniqueId(Map<String, Integer> ids, String name) {
```
apiri: I see what is happening here but wonder if hashing might be a better alternative.  While this does avoid collisions that could result, could keep things simpler given the requirement for unique names currently.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
```diff
@@ -546,6 +550,37 @@ protected static void addConnection(final Element parentElement, ConnectionSchem
...
+        String baseId = name.replaceAll("[^A-Za-z0-9_]", "_");
+        String id = baseId;
+        Integer idNum = ids.get(baseId);
+        while (ids.containsKey(id)) {
+            id = baseId + "_" + idNum++;
+        }
+        if (id != baseId) {
+            ids.put(baseId, idNum);
+        }
+        ids.put(id, 2);
```
apiri: if we opt to keep the current approach, not sure I follow this.  Was this a remnant of testing?

apiri:
Changes look like they would resolve the issue in question but would be interested to hear your thoughts on opting for hashing what is their currently.  Thanks!

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
brosander: @apiri so I may have overthought this a little bit.  

The initial approach I just tracked used names and kept incrementing a suffix.  It bothered me a little that that would be n^2 with respect to duplicate names, the above should avoid that pitfall by tracking the next suffix available for every prefix.  That way the following wouldn't collide:

test -> test
test -> test_2
test_2 -> test_2_2

apiri: Ah, I see.  Wasn't tracking and certainly didn't have that pop into mind.

apiri:
@brosander all are fair points and definitely was looking at collisions as a risk.  my primary focus was looking to relegate the name to nothing more than a variable to get it in a similar space as how NiFi handles these things.  To that end, maybe we introduce an id that is required to be unique.  From our curent toolkit, we could just lift the uuid that NiFi creates.  In the case that someone is creating a flow from scratch/handwriting, they would be responsible for generating a unique one.  This should alleviate the woes we have run into, mapping more closely to NiFi's component "namespace".

This will require thoughts and handling of how to take existing flows that may not have an ID as generated by the previous toolkit.

We should additionally maybe start including some metadata about the YAML format such that we can make decisions in transform/loading that allow us to evolve.

brosander:
@apiri thinking of adding version tag to yml, think we should stick with MiNiFi versioning for it?

e.g. use 0.1.0 for current master?

apiri:
I think incremental is probably most straightforward and would likely just call this version 2 if that seems fair.

JPercivall:
Hey @brosander, I started taking a look at the PR and there is a lot of great work here. When looking through the PR I see that the IDs are only for connections. I'm thinking it would make more sense to just blanket add them to everything instead of just saying "Oh those are just for connections". I believe it would make more sense to the user and would let us address MINIFI-106 too. 

Thoughts @apiri?

Also a comment on the PR, don't forget to add documentation about config schema changes to the "System_Admin_Guide.md" in the docs module.

apiri:
Sounds fair.  I wonder if it makes sense to treat the IDs as optional properties overall where we use kind of the current way to minimize pain and allow custom specification for advanced cases.  I realize the file isn't the end desire for configuration, but in the interim, making the format as friendly as possible would be helpful.

JPercivall:
@apiri makes sense to me

brosander:
Closing in favor of #33

PR #33, opened 2016-09-15
Minifi 108

brosander:


minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConfigSchema.java
```diff
@@ -42,10 +44,14 @@
  *
  */
 public class ConfigSchema extends BaseSchema {
-    public static final String FOUND_THE_FOLLOWING_DUPLICATE_PROCESSOR_NAMES = "Found the following duplicate processor names: ";
-    public static final String FOUND_THE_FOLLOWING_DUPLICATE_CONNECTION_NAMES = "Found the following duplicate connection names: ";
+    public static final String FOUND_THE_FOLLOWING_DUPLICATE_PROCESSOR_IDS = "Found the following duplicate processor ids: ";
+    public static final String FOUND_THE_FOLLOWING_DUPLICATE_CONNECTION_IDS = "Found the following duplicate connection ids: ";
     public static final String FOUND_THE_FOLLOWING_DUPLICATE_REMOTE_PROCESSING_GROUP_NAMES = "Found the following duplicate remote processing group names: ";
+    public static final String FOUND_THE_FOLLOWING_DUPLICATE_REMOTE_INPUT_PORT_IDS = "Found the following duplicate remote input port ids: ";
+    public static final String FOUND_THE_FOLLOWING_DUPLICATE_IDS = "Found the following ids that occur both in Processors and Remote Input Ports: ";
```
JPercivall: This brings up a good point that all the ids within a PG (ie. all the ids we have) will have to be unique. So will need to add a check including connections and RPGs themselves.

minifi-commons/minifi-commons-schema/src/test/java/org/apache/nifi/minifi/commons/schema/serialization/SchemaLoaderTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class SchemaLoaderTest {
```
JPercivall: I think this could benefit a bit by adding two more tests, ones with a schemas that explicitly have version 1 and 2 listed (these all use "config-minimal.yml" which has no version key).

minifi-toolkit/minifi-toolkit-configuration/src/test/resources/CsvToJson.yml
```diff
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+MiNiFi Config Version: 2
```
JPercivall: The config files in the toolkit were updated but not the ones in minifi-bootstrap or the default one that gets bundled (in minifi-resources). Could you update those too?

Kinda sucks that we have test yml files in multiple locations that will need to be updated each time there is a change to the schema. Any thoughts to remedy this/if you think it's worth it?

minifi-toolkit/minifi-toolkit-configuration/src/test/resources/InvokeHttpMiNiFiTemplateTest.yml
```diff
@@ -255,8 +273,8 @@ Remote Processing Groups:
   timeout: 30 sec
```
brosander: The input ports already have ids, which is what should matter.  I think we can add some validation though.

minifi-toolkit/minifi-toolkit-configuration/src/test/resources/StressTestFramework.yml
```diff
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+MiNiFi Config Version: 2
```
brosander: Will do, ended up using version 1 for this one though so that ids are optional as long as names are unique.

minifi-toolkit/minifi-toolkit-configuration/src/test/resources/CsvToJson.yml
brosander: We could move them all to a test/resources folder and use it as a dependency.

minifi-commons/minifi-commons-schema/src/test/java/org/apache/nifi/minifi/commons/schema/serialization/SchemaLoaderTest.java
brosander: Since making the ids optional, we don't have 2 distinct versions anymore.  Leaving versioning part of SchemaLoader in place though

brosander:
@JPercivall I wound up reworking quite a bit to account for source id, destination id in connections, allow for names instead in processor, connection, input port

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConfigSchema.java
```diff
@@ -198,4 +305,31 @@ public ComponentStatusRepositorySchema getComponentStatusRepositoryProperties()
...
+     * @return a unique filesystem-friendly id
+     */
+    protected static String getUniqueId(Map<String, Integer> ids, String name) {
+        String baseId = StringUtil.isNullOrEmpty(name) ? EMPTY_NAME : ID_REPLACE_PATTERN.matcher(name).replaceAll("_");
+        String id = baseId;
+        Integer idNum = ids.get(baseId);
+        while (ids.containsKey(id)) {
+            id = baseId + "_" + idNum++;
+        }
+        if (id != baseId) {
```
JPercivall: Both "id" and "baseId" are Strings, are you comparing using "!=" instead of "!id.equals(..." on purpose?

minifi-docs/src/main/markdown/System_Admin_Guide.md
```diff
@@ -174,10 +174,11 @@ for example "10 sec" or "10 MB", not simply "10".
 
 The first section of config.yml is for naming and commenting on the file.
 
- Property | Description
- -------- | ---
-name      | The name of the file.
-comment   | A comment describing the usage of this config file.
+ Property             | Description
+--------------------- | -----------
+MiNiFi Config Version | The version of the configuration file.
```
JPercivall: This should include a little more information for the user to understand what they should put and why. Like that there is currently only 1 version and the default value will be version 1.

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConfigSchema.java
```diff
@@ -96,51 +100,154 @@ public ConfigSchema(Map map) {
...
+
+        connections.stream().filter(connection -> StringUtil.isNullOrEmpty(connection.getSourceId())).forEach(connection -> {
+            String sourceName = connection.getSourceName();
+            if (remoteInputPortIds.contains(sourceName)) {
+                connection.setSourceId(sourceName);
+            } else {
+                if (duplicateProcessorNames.contains(sourceName)) {
+                    problematicDuplicateNames.add(sourceName);
+                }
+                connection.setSourceId(processorNameToIdMap.get(sourceName));
```
JPercivall: This (and the corresponding line for dest names) should check if source name is actually in the list of processors (it's already checking if it's in the list of remote input ports). I tested a flow where I didn't put the right name for "source name" and got this confusing error message: 

Exception in thread "main" java.io.IOException: Unable to successfully transform the provided configuration
	at org.apache.nifi.minifi.bootstrap.RunMiNiFi.performTransformation(RunMiNiFi.java:1543)
	at org.apache.nifi.minifi.bootstrap.RunMiNiFi.start(RunMiNiFi.java:1083)
	at org.apache.nifi.minifi.bootstrap.RunMiNiFi.main(RunMiNiFi.java:215)
Caused by: org.apache.nifi.minifi.bootstrap.exception.InvalidConfigurationException: Failed to transform config file due to:['source id' in section 'Connections' because it was not found and it is required]
	at org.apache.nifi.minifi.bootstrap.util.ConfigTransformer.transformConfigFile(ConfigTransformer.java:92)
	at org.apache.nifi.minifi.bootstrap.RunMiNiFi.performTransformation(RunMiNiFi.java:1539)
	... 2 more

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConfigSchema.java
brosander: @JPercivall I'm aware that most of the time this signals a bug, the assignment of id to baseId is only a few lines up, I'm just trying to see if it was reassigned in the while loop, meaning the reference equals should (and seems to based on testing) do the trick.

I can change to a boolean flag if you think it would make things clearer.

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConfigSchema.java
brosander: @JPercivall good point, I will try to add a more specific message for that case.

minifi-docs/src/main/markdown/System_Admin_Guide.md
brosander: @JPercivall good point, I will try to clarify

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConfigSchema.java
JPercivall: Totally down with "!=" for this case.

Lets just comment why it's a valid case so people aren't confused later on.

brosander:
@JPercivall I believe I've addressed your feedback.  I've also run through https://github.com/brosander/minifi-testkit with it and tested swapping on a windows vm.

JPercivall:
+1

Visually verified code and did a contrib check build. Went through many different configurations to check that id/name worked correctly and everything worked as expected. Thanks @brosander, I will merge it in.

PR #35, opened 2016-09-21
MINIFI-59 Removing exclude statement for Quartz to fix CRON support

JPercivall:


apiri:
reviewing

apiri:
Created a flow template using Cron, converted, and ran and all looks good.  Will merge.  Thanks!

PR #37, opened 2016-09-21
MINIFI-110 Adding ' ' (space) to the list of characters that will cau…

JPercivall:
…se parsing errors when using the 'flowStatus' command


apiri:
Looks fine, will merge.

PR #39, opened 2016-09-23
MINIFI-81 Adding a check for a valid command to the config.sh toolkit…

JPercivall:
… script

### DIFF
--- a/minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
+++ b/minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
@@ -85,8 +85,6 @@ public class ConfigMain {
     public static void printValidateUsage() {
         System.out.println("Validate Usage:");
         System.out.println();
-        System.out.print("java ");
-        System.out.print(ConfigMain.class.getCanonicalName());
         System.out.println(" validate INPUT_FILE");
         System.out.println();
     }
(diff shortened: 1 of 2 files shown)


apiri:
reviewing

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
```diff
@@ -85,8 +85,6 @@ public static void main(String[] args) {
     public static void printValidateUsage() {
         System.out.println("Validate Usage:");
         System.out.println();
-        System.out.print("java ");
```
apiri: Should this and the other java invocation remain given that the interpretation of arguments is handled by the script when invoked from the script?  This is valid in the context the class is run by itself.

apiri:
Everything looks good.  Just wondering if we should keep usage for the main class such that it is applicable if run independently of the script.  If not, would think we should perhaps just nix those two methods entirely.

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
JPercivall: Hey @apiri, I chose to remove them because they can lead to very confusing errors messages for user who ran it using the scripts. If they happen to mis-type a file location not only would they see the error saying the file doesn't exist but also this "java <class name> ..." line that was quite confusing.

Also, I see this being run using the script in >99% of the cases.

apiri: fair enough, will merge

PR #40, opened 2016-09-30
MINIFI-46 - Supporting multiple relationships for MiNiFi connections

brosander:


JPercivall:
Reviewing

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConfigSchema.java
```diff
@@ -53,7 +53,7 @@
     public static final String FOUND_THE_FOLLOWING_DUPLICATE_REMOTE_INPUT_PORT_IDS = "Found the following duplicate remote input port ids: ";
     public static final String FOUND_THE_FOLLOWING_DUPLICATE_IDS = "Found the following ids that occur both in Processors and Remote Input Ports: ";
     public static final String CANNOT_LOOK_UP_PROCESSOR_ID_FROM_PROCESSOR_NAME_DUE_TO_DUPLICATE_PROCESSOR_NAMES = "Cannot look up Processor id from Processor name due to duplicate Processor names: ";
-    public static final int CONFIG_VERSION = 1;
+    public static final int CONFIG_VERSION = 2;
```
JPercivall: I was doing some testing around schemas and their version and I think we may overlooked validation using the toolkit. I did this test: I have a config with a config_version "1" but it has multiple relationships for a connection. When I run it through the toolkit validator it will pass because that will work in the current implementation but if we took that config file to a MiNiFi 0.0.1 instance (that would be expected to be able to run a version "1" config) it will fail. 

As we go forward with Command and Control (with multiple versions of MiNiFi in the wild) I think we'll need to be strict with validation of the config version in the toolkit but when the MiNiFi instance is actually trying to run it, be as lenient as practical. So use the SchemaLoader to transition any keys/formats that were supported in a prior version but not in the future version (ie. deprecated things). 

I think this should allow us a go way forward in terms of maintaining backwards compatibility while keeping expected functionality. This would be outside the scope of this PR/ticket though.

brosander: @JPercivall that's a really good point, I'll write up another issue for it.

brosander:
@JPercivall I added the validation logic you suggested.  While I was in there I also added unit tests for most of the ConnectionSchema functionality.

JPercivall:
Only comment is that the "only" in the validation issue for ConnectionSchema should be capitalized. I will amend that to your commit.

+1

Visually verified code and did a contrib check build. Tested multiple different configs with one+ relationships per connection. Looks good @brosander, thanks for the contribution. I will merge it in.

PR #42, opened 2016-10-12
MINIFI-120 Fixing 'class not found' issues when using the Rest Change…

JPercivall:
… Notifier. Also added all the jetty and javax modules that NiFi lists as 'provided' in its pom to the minifi-assembly pom in order to ensure NAR reusability.


apiri:
reviewing

apiri:
Looks fair.  Was able to verify issue and correction.  Will get this merged in, thanks for finding and fixing.

PR #45, opened 2016-10-18
MINIFI-117 - Maintainable Configuration Versioning

brosander:


brosander:
Made V1 schema classes for schemas that have changed since 0.0.1.

V1 schemas implement ConvertableSchema meaning they can self-validate according to their version's rules and be converted to current schema.

Current Schemas implement WritableSchema meaning they can convert themselves to a Map for writing to a file (used by toolkit).

Moved backwards compatibility logic to the convert() phase.

Made config validation 2 phase:
1. Parse yaml into its config schema version, check for errors.
2. Convert to current version, check for errors.

Added tests cases verifying the conversion works as expected.

JPercivall:
Reviewing

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConnectionSchema.java
```diff
@@ -60,30 +53,11 @@ public ConnectionSchema(Map map) {
         super(map, CONNECTIONS_KEY);
 
         sourceId = getOptionalKeyAsType(map, SOURCE_ID_KEY, String.class, CONNECTIONS_KEY, "");
```
JPercivall: I believe this should now be required instead of optional (since there is no id vs name confusion)

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConnectionSchema.java
```diff
@@ -60,30 +53,11 @@ public ConnectionSchema(Map map) {
...
-                sourceRelationshipNames = getRequiredKeyAsType(map, SOURCE_RELATIONSHIP_NAMES_KEY, List.class, CONNECTIONS_KEY);
-            } else {
-                sourceRelationshipNames = new ArrayList<>(Arrays.asList(sourceRelationshipName));
-            }
+        sourceRelationshipNames = getOptionalKeyAsType(map, SOURCE_RELATIONSHIP_NAMES_KEY, List.class, CONNECTIONS_KEY, new ArrayList<>());
+        if (sourceRelationshipNames.isEmpty()) {
+            addValidationIssue("Expected at least one value in " + SOURCE_RELATIONSHIP_NAMES_KEY + " for " + CONNECTIONS_KEY + " " + getName());
         }
-
         destinationId = getOptionalKeyAsType(map, DESTINATION_ID_KEY, String.class, CONNECTIONS_KEY, "");
```
JPercivall: Same comment as on source, I think this should be "required".

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/serialization/SchemaLoader.java
```diff
@@ -30,13 +32,13 @@
...
 
 public class SchemaLoader {
-    private static final Map<String, Function<Map, ConfigSchema>> configSchemaFactories = initConfigSchemaFactories();
+    private static final Map<String, Function<Map, ConvertableSchema<ConfigSchema>>> configSchemaFactories = initConfigSchemaFactories();
 
-    private static Map<String, Function<Map, ConfigSchema>> initConfigSchemaFactories() {
-        Map<String, Function<Map, ConfigSchema>> result = new HashMap<>();
-        result.put(String.valueOf((Object)null), ConfigSchema::new);
-        result.put("", ConfigSchema::new);
-        result.put("1", ConfigSchema::new);
```
JPercivall: The changes in the Config schemas need to be documented. Could you add a section to the admin guide (the Config Schema portion) that explains the versioning scheme as well as what changed in v2? 

I am suggesting to put it in the docs instead of a migration guidance for a release because working with multiple different version of MiNiFi and configs will be an everyday task (esp when C&C is added).

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
```diff
@@ -96,13 +99,21 @@ public int validate(String[] args) {
...
             try {
-                ConfigSchema configSchema = SchemaLoader.loadConfigSchemaFromYaml(inputStream);
+                ConvertableSchema<ConfigSchema> configSchema = SchemaLoader.loadConvertableSchemaFromYaml(inputStream);
                 if (!configSchema.isValid()) {
                     configSchema.getValidationIssues().forEach(s -> System.out.println(s));
                     System.out.println();
                     return ERR_INVALID_CONFIG;
                 } else {
-                    System.out.println(NO_VALIDATION_ERRORS_FOUND_IN_TEMPLATE);
+                    ConfigSchema currentSchema = configSchema.convert();
```
brosander: I think it would, good call :)

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConnectionSchema.java
brosander: It is treated as required in the getValidationIssues implementation.  I needed a chance to set it when upconverting before the validation occurs.  I could change that to map manipulation before instantiating the ConnectionSchema but that felt worse.

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/serialization/SchemaLoader.java
brosander: Good point, I'll take a stab at updating the doc

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConnectionSchema.java
JPercivall: Ah I see, could you just add a comment above these two "optional" calls explaining that?

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/ConfigSchema.java
```diff
@@ -142,128 +125,10 @@ public ConfigSchema(Map map) {
...
-                        }
-                    }
-                });
-
-        if (problematicDuplicateNames.size() > 0) {
-            addValidationIssue(CANNOT_LOOK_UP_PROCESSOR_ID_FROM_PROCESSOR_NAME_DUE_TO_DUPLICATE_PROCESSOR_NAMES
-                    + problematicDuplicateNames.stream().collect(Collectors.joining(", ")));
-        }
-        if (missingProcessorNames.size() > 0) {
-            addValidationIssue(CONNECTIONS_REFER_TO_PROCESSOR_NAMES_THAT_DONT_EXIST + missingProcessorNames.stream().sorted().collect(Collectors.joining(", ")));
```
brosander: I agree we should validate that the connections refer to existing ids, will add that

brosander:
@JPercivall I believe I've addressed your concerns, I also added a command to the toolkit to upgrade an older yml file to the current version.

Please let me know what you think.

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
```diff
@@ -96,31 +102,39 @@ public int validate(String[] args) {
         }
         try (InputStream inputStream = pathInputStreamFactory.create(args[1])) {
             try {
-                ConfigSchema configSchema = SchemaLoader.loadConfigSchemaFromYaml(inputStream);
+                ConvertableSchema<ConfigSchema> configSchema = SchemaLoader.loadConvertableSchemaFromYaml(inputStream);
+                boolean valid = true;
                 if (!configSchema.isValid()) {
+                    System.out.println(FOUND_THE_FOLLOWING_ERRORS_WHEN_PARSING_THE_TEMPLATE_ACCORDING_TO_ITS_VERSION);
```
brosander: Good point, I'll add that.

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
```diff
@@ -96,31 +102,39 @@ public int validate(String[] args) {
...
                     System.out.println();
-                    return ERR_INVALID_CONFIG;
-                } else {
+                    valid = false;
+                    configSchema.clearValidationIssues();
+                }
+
+                ConfigSchema currentSchema = configSchema.convert();
+                if (!currentSchema.isValid()) {
+                    System.out.println(WHEN_CONVERTING_TO_LATEST_VERSION_FOUND_THE_FOLLOWING_ERRORS);
```
brosander: Adding

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
```diff
@@ -235,62 +252,127 @@ private static void addRemoteProcessGroupPortDTOs(Map<String, String> connectabl
...
+    public int upgrade(String[] args) {
+        if (args.length != 3) {
+            printUgradeUsage();
+            return ERR_INVALID_ARGS;
+        }
+
+        ConfigSchema configSchema = null;
+        try (InputStream inputStream = pathInputStreamFactory.create(args[1])) {
+            try {
+                configSchema = SchemaLoader.loadConfigSchemaFromYaml(inputStream);
```
brosander: @JPercivall for the latest schema, the convert will just return this.  I can add extra logic to detect that case and do nothing but it seems like that's harmless behavior and could be used to round-trip and/or reformat the file.

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
```diff
@@ -235,62 +252,127 @@ private static void addRemoteProcessGroupPortDTOs(Map<String, String> connectabl
...
+            } catch (IOException e) {
+                return handleErrorSavingCofiguration(e);
+            }
+        } catch (FileNotFoundException e) {
+            return handleErrorOpeningOutput(args[2], ConfigMain::printUgradeUsage, e);
+        } catch (IOException e) {
+            handleErrorClosingOutput(e);
+        }
+
+        return SUCCESS;
```
brosander: Good point, adding

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
```diff
@@ -96,31 +102,39 @@ public int validate(String[] args) {
...
+                    System.out.println(FOUND_THE_FOLLOWING_ERRORS_WHEN_PARSING_THE_TEMPLATE_ACCORDING_TO_ITS_VERSION);
                     configSchema.getValidationIssues().forEach(s -> System.out.println(s));
                     System.out.println();
-                    return ERR_INVALID_CONFIG;
-                } else {
+                    valid = false;
+                    configSchema.clearValidationIssues();
+                }
+
+                ConfigSchema currentSchema = configSchema.convert();
```
brosander: @JPercivall I'll take a stab at it. I'm not sure if it will be more or less confusing though, please let me know what you think.

brosander:
@JPercivall I believe I've addressed all of your comments except for the config upgrade version check one.  Please let me know if my reply is satisfactory and/or if you see more things that should be changed/fixed.

minifi-commons/minifi-commons-schema/src/main/java/org/apache/nifi/minifi/commons/schema/serialization/SchemaLoader.java
```diff
@@ -78,6 +80,27 @@ public static ConfigSchema loadConfigSchemaFromYaml(Map<String, Object> yamlAsMa
 
     public static ConvertableSchema<ConfigSchema> loadConvertableSchemaFromYaml(Map<String, Object> yamlAsMap) throws SchemaLoaderException {
         String version = String.valueOf(yamlAsMap.get(ConfigSchema.VERSION));
+        if (StringUtil.isNullOrEmpty(version) || String.valueOf((Object) null).equals(version)) {
```
brosander: @JPercivall sorry, I thought you were wanting behavior like that in regards to [your comment above.](https://github.com/apache/nifi-minifi/pull/45#discussion_r85137468)

I must've misread the comment, now it's looking more like it may have been automatically addressed by the other validation change which clears the errors before converting.

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
```diff
@@ -336,6 +329,16 @@ public int transform(String[] args) {
         return SUCCESS;
     }
 
+    protected void validateAndPrintIssues(ConfigSchema configSchema) {
+        if (!configSchema.isValid()) {
+            System.out.println("There are validation errors with the template, still outputting YAML but it will need to be edited.");
+            configSchema.getValidationIssues().forEach(System.out::println);
```
JPercivall: I don't think the validation issues are getting cleared properly along the way:

screen shot 2016-10-28 at 4 08 01 pm

brosander: @JPercivall the first printout looks right to me.  You're missing the name in v1 which gets cleared.  Then when it tries to create the v2 schema, it sees the id missing.

I can do similar logic with the upgrade command as well though.

brosander:
@JPercivall went back to old schema loading logic, filtering out non-user-friendly versions when printing error

Also extracted validation printing logic out to a function and call it from both validate and upgrade.

JPercivall:
+1

Visually verified code and did a contrib check build. Verified both the toolkit and MiNiFi assembly using multiple different versions of the config files. All worked as expected. Thanks @brosander, I will merge this in.

PR #51, opened 2016-11-07
MINIFI-36 Refactoring config change notifiers and adding Pull config …

JPercivall:
…change notifier

### For all changes:
- [X] Is there a JIRA ticket associated with this PR? Is it referenced 
     in the commit message?

- [X] Does your PR title start with MINIFI-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [X] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [X] Is your initial contribution a single, squashed commit?

### For code changes:
- [X] Have you ensured that the full suite of tests is executed via mvn -Pcontrib-check clean install at the root nifi-minifi folder?
- [X] Have you written or updated unit tests to verify your changes?
- [X] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? 
- [X] If applicable, have you updated the LICENSE file, including the main LICENSE file under minifi-assembly?
- [X] If applicable, have you updated the NOTICE file, including the main NOTICE file found under minifi-assembly?

- [X] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:


brosander:
Reviewing

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
```diff
@@ -143,11 +145,18 @@
...
     private Set<PeriodicStatusReporter> periodicStatusReporters;
 
+    private ConfigurationChangeCoordinator changeCoordinator;
     private MiNiFiConfigurationChangeListener changeListener;
 
+    private volatile ByteBuffer currentConfigFile;
+
+    @Override
+    public ByteBuffer getConfigFile() {
+        return currentConfigFile;
```
brosander: Can we make a defensive copy or provide an immutable value here?  I'm worried about multiple callers getting a stateful object.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/configuration/ConfigurationChangeCoordinator.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /**
+     * Provide the mechanism by which listeners are notified
+     */
+    public Collection<ListenerHandleResult> notifyListeners(ByteBuffer newConfig) {
+        logger.info("Notifying Listeners of a change");
+
+        Collection<ListenerHandleResult> listenerHandleResults = new ArrayList<>(configurationChangeListeners.size());
+        for (final ConfigurationChangeListener listener : getChangeListeners()) {
```
brosander: Are we anticipating listener registration on different threads?  This could bomb if listener is added during iteration.

minifi-docs/src/main/markdown/System_Admin_Guide.md
```diff
@@ -21,6 +21,102 @@
 
 [MiNiFi Homepage](https://nifi.apache.org/minifi/index.html)
 
+# Automatic Warm-Redeploy
+
+When many MiNiFi agents running on the edge, it may not be possible to manually stop, edit the config.yml and then restart every one every time their configuration needs to change. The Config Change Coordinator and it's Ingestors were designed to automatically redeploy in response to a configuration update.
```
brosander: nitpick: it's -> its

minifi-docs/src/main/markdown/System_Admin_Guide.md
```diff
@@ -21,6 +21,102 @@
...
+ - RestChangeIngestor
+ - PullHttpChangeIngestor
+
+After a new configuration has been pulled/received the Ingestors use a Differentiator in order to determine if the currently running config is different than the new config. Which Differentiator is used, is configurable for each Ingestor. Currently there is only one Differentiator:
+
+ - WholeConfigDifferentiator: Compares the entire new config with the currently running one, byte for byte.
+
+After a new config is determined to be new, the MiNiFi agent will attempt to restart. The bootstrap first saves the old config into a swap file. The bootstrap monitors the agent as it restarts and if it fails it will roll back to the old config. If it succeeds then the swap file will be deleted and the agent will start processing using the new config.
+
+Note: Data left in connections when the agent attempts to restart will either mapped to a connection with the same ID in the new config, or orphaned and deleted.
```
brosander: either mapped -> either be mapped

minifi-bootstrap/src/test/java/org/apache/nifi/minifi/bootstrap/configuration/mocks/MockDifferentiator.java
```diff
@@ -0,0 +1,47 @@
...
+
+package org.apache.nifi.minifi.bootstrap.configuration.mocks;
+
+import org.apache.nifi.minifi.bootstrap.ConfigurationFileHolder;
+import org.apache.nifi.minifi.bootstrap.configuration.differentiators.interfaces.Differentiator;
+
+import java.io.IOException;
+import java.util.Properties;
+
+public class MockDifferentiator<T> implements Differentiator<T> {
```
brosander: Can we use mockito and a when -> thenReturn for this?

brosander:
@JPercivall I made a first pass over the changes and like it overall.  About to start functional testing.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
JPercivall: That's a fair point. How about an AtomicReference to a read only version of the buffer?

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/configuration/ConfigurationChangeCoordinator.java
JPercivall: In the current implementation we only have one listener (RunMiNiFi) and we call init->register->start all in RunMiNiFi. When we do have more I would imagine we would know exactly what listeners to register when we init. 

So I think what we should do is remove the "registerListener" method and instead make "init" take a list of Listeners too. Thoughts?

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
brosander: Is the position in the buffer read-only?

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/configuration/ConfigurationChangeCoordinator.java
brosander: Works for me

minifi-bootstrap/src/test/java/org/apache/nifi/minifi/bootstrap/configuration/mocks/MockConfigurationFileHolder.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.nifi.minifi.bootstrap.configuration.mocks;
+
+import org.apache.nifi.minifi.bootstrap.ConfigurationFileHolder;
+
+import java.nio.ByteBuffer;
+
+public class MockConfigurationFileHolder implements ConfigurationFileHolder {
```
JPercivall: Definitely can and we can use it for all the classes in my mock folder. Totally should have done that from the beginning, removing all of the "Mock" classes now.

minifi-bootstrap/src/test/java/org/apache/nifi/minifi/bootstrap/configuration/mocks/MockDifferentiator.java
JPercivall: As stated on the other comment, removing all mock classes in my test folder

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
JPercivall: Here is the docs for ByteBuffer's "asReadyOnlyBuffer". The position is independent for each but share the same content (and prevent modification to the content). 

One thing to note, the "currentConfigFile" won't ever be updated by writing to it (reference is updated when a new config is transformed). So we can make the underlying "currentConfigFile" read only too (effectively making it immutable).

https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#asReadOnlyBuffer--

brosander: Sounds good

brosander:
@JPercivall the FileChangeIngestor should probably fire once on startup to pick up changes that were made before startup.

brosander:
@JPercivall using PullHttpChangeIngestor with 
```
python -m SimpleHTTPServer
```

1. It would be nice to be able to specify the path.  That way you can more easily use an existing webserver that is already hosting other content.
2. It seems to always detect a changed file (I tried with every 30 seconds) whether or not the file has changed.
3. Setting the polling interval to be faster than a restart (e.g. 1 second) causes strange behavior, null pointers, etc

JPercivall:
@brosander expanding on the "FileChangeIngestor fire once first" point, I thought about it a bit and I think the best course of action would be to optionally (configurable in bootstrap.conf) check each ingestor before starting up (would currently apply to File and PullHTTP). That way before starting up, MiNiFi can be sure it has the most up to date Config and won't have to restart with the first invocation of a the change notifiers.

That said I think it should be a new ticket since it would require a bit of refactoring of RunMiNiFi and that work is not explicitly related to this current ticket.

JPercivall:
Hey @brosander, answering you comment:

1: I agree, will add
2: You saw it successfully restart every time? I am not able to reproduce
3: Yeah that's gonna suck a bit, I can look into what I can though. Also there is some brittle logic leftover in NiFi in the main while loop which can lead to some problems that doesn't help

JPercivall:
@brosander just pushed out a new commit addressing your feedback (except for number 2 above).

brosander:
@JPercivall now I'm getting the following when copying a file into nifi.minifi.notifier.ingestors.file.config.path with the FileChangeIngestor.

This is after a fresh unzip and only changing the ingestor and the path.

```
2016-11-10 09:50:55,150 INFO [File Change Notifier Thread] o.a.n.m.b.c.ConfigurationChangeCoordinator Notifying Listeners of a change                                                 2016-11-10 09:50:55,150 INFO [File Change Notifier Thread] o.apache.nifi.minifi.bootstrap.RunMiNiFi Received notification of a change                                                 2016-11-10 09:50:55,153 ERROR [File Change Notifier Thread] o.apache.nifi.minifi.bootstrap.RunMiNiFi Unable to carry out reloading of configuration on receipt of notification event  org.apache.nifi.minifi.bootstrap.configuration.ConfigurationChangeException: Instance is already reloading                                                                                    at org.apache.nifi.minifi.bootstrap.RunMiNiFi$MiNiFiConfigurationChangeListener.handleChange(RunMiNiFi.java:1518) ~[minifi-bootstrap-0.1.0-SNAPSHOT.jar:0.1.0-SNAPSHOT]               at org.apache.nifi.minifi.bootstrap.configuration.ConfigurationChangeCoordinator.notifyListeners(ConfigurationChangeCoordinator.java:96) [minifi-bootstrap-0.1.0-SNAPSHOT.jar:0.1.0-SNAPSHOT]                                                                                                                                                                               at org.apache.nifi.minifi.bootstrap.configuration.ingestors.FileChangeIngestor.run(FileChangeIngestor.java:147) [minifi-bootstrap-0.1.0-SNAPSHOT.jar:0.1.0-SNAPSHOT]                  at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [na:1.8.0_102]                                                                                             at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) [na:1.8.0_102]                                                                                                    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) [na:1.8.0_102]                                               at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) [na:1.8.0_102]                                                      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [na:1.8.0_102]                                                                                     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [na:1.8.0_102]                                                                                     at java.lang.Thread.run(Thread.java:745) [na:1.8.0_102]                                                                                                                       2016-11-10 09:50:55,153 INFO [File Change Notifier Thread] o.a.n.m.b.c.ConfigurationChangeCoordinator Listener notification result:MiNiFiConfigurationChangeListener FAILED to handle the configuration change due to: 'Instance is already reloading'
```

JPercivall:
That's what I get for not testing properly. I had to stop coding and wanted to address all your feedback before pushing a PR and I messed up addressing your number 3. Sorry @brosander, I know how frustrating it is to go to review and it doesn't work.

I will fix it now.

JPercivall:
@brosander just pushed the fix. I tested it by having all 3 notifiers running at the same time and sending changes to all of them many times. I didn't see any issues. Again, sorry for rushing.

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
```diff
@@ -1600,9 +1622,13 @@ private void restartInstance() throws IOException {
...
-    private static void performTransformation(InputStream configIs, String configDestinationPath) throws ConfigurationChangeException, IOException {
-        try {
-            ConfigTransformer.transformConfigFile(configIs, configDestinationPath);
+    private static ByteBuffer performTransformation(InputStream configIs, String configDestinationPath) throws ConfigurationChangeException, IOException {
+        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+                TeeInputStream teeInputStream = new TeeInputStream(configIs, byteArrayOutputStream)) {
+
+            ConfigTransformer.transformConfigFile(teeInputStream, configDestinationPath);
+
+            return ByteBuffer.wrap(byteArrayOutputStream.getUnderlyingBuffer());
```
brosander: @JPercivall I think this is what was causing my constant restarting to happen.  If the underlying buffer isn't full, it will result in a bunch of null values at the end in the transformed config.

JPercivall:
@brosander I've pushed the change fixing the constant restart

brosander:
+1 committed

PR #57, opened 2016-11-28
MINIFI-137 Adding option to RPGs to transmit via HTTP. Also corrected…

JPercivall:
… a typo that had RPGs as 'Remote Processing Groups' instead of 'Remote Process Groups' (extra 'ing').


brosander:
Reviewing

minifi-toolkit/minifi-toolkit-configuration/src/test/resources/ProcessGroupsAndRemoteProcessGroups.yml
```diff
@@ -173,13 +173,14 @@ Process Groups:
       max work queue data size: 1 GB
       flowfile expiration: 0 sec
       queue prioritizer class: ''
-    Remote Processing Groups:
+    Remote Process Groups:
```
brosander: Definitely good to update expected output for test ymls, where possible though it would be nice to add a -v1.yml so we validate the upconvert works as expected.  (ex: DecompressionCircularFlow-v1.yml vs DecompressionCircularFlow.yml)

It should be automagically picked up by ConfigMainTest.testV1YmlIfPresent() if named via the convention.

minifi-commons/minifi-commons-schema/src/test/java/org/apache/nifi/minifi/commons/schema/RemoteProcessGroupSchemaTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Test
+    public void testTransportProtocol() {
+        Map<String, Object> map = new HashMap<>();
+        map.put(CommonPropertyKeys.INPUT_PORTS_KEY, Arrays.asList(createPortSchema("f94d2469-39f8-4f07-a0d8-acd9396f639e", "testName", ConfigSchema.TOP_LEVEL_NAME).toMap()));
+        map.put(RemoteProcessGroupSchema.URL_KEY, "http://localhost:8080/nifi");
+        map.put(CommonPropertyKeys.ID_KEY, "a58d2fab-7efe-4cb7-8224-12a60bd8003d");
+        map.put(RemoteProcessGroupSchema.TRANSPORT_PROTOCOL_KEY, "not valid");
+        validateIssuesNumMatches(1, new RemoteProcessGroupSchema(map));
+
```
brosander: Might make sense here to assert the transport protocol getter returns the correct value for raw and http keys.

brosander:
+1 conditional on above comments (testing nitpicks) being addressed and/or discussed.

Tested http and raw rpg after transforming [InvokeHttpMiNiFiTemplateTest](https://github.com/brosander/minifi-testkit/blob/master/templates/InvokeHttpMiNiFiTemplateTest.xml)

Ran through all templates in [my testkit](https://github.com/brosander/minifi-testkit/tree/37058364155ee7679ebd424b5cc16eda06b85cd7/templates)

Tried malforming the RPG field a few ways and validating/upgrading the config.yml.

JPercivall:
@brosander just pushed out a new commit addressing your concerns. I couldn't add a v1 version of "ProcessGroupsAndRemoteProcessGroups" since it uses Process Groups and a v1 config would not be able to represent it. I did add another test flow that is a simple TailFile to RPG which has a v1 version.

brosander:
+1, merging

PR #67, opened 2016-12-20
MINIFI-144 Bumping logback version to 1.1.8

JPercivall:


apiri:
Thanks for updating this.  Looks like the size cap is working more or less as described.

Our assembly NOTICE lists the versions we use, could you update those?

Could you also update the logback.xml to remove reference to the issue that was resolved in this version?  

Otherwise it looks good to go.

JPercivall:
Only missed 2/3's of the places I needed to change, haha. Thanks for catching them and I have updated the PR.

apiri:
1 out of 3 ain't bad!  Thanks for adjusting!  Will merge.

PR #105, opened 2017-12-13
MINIFI-418: Added docs on Windows service for 64bit machine

ijokarumawak:

### For all changes:
     in the commit message?

- [ ] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] Have you ensured that the full suite of tests is executed via mvn -Pcontrib-check clean install at the root nifi-minifi folder?

- [x] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:

### DIFF
--- a/minifi-docs/src/main/markdown/System_Admin_Guide.md
+++ b/minifi-docs/src/main/markdown/System_Admin_Guide.md
@@ -656,11 +656,13 @@ NiFi Properties Overrides:
 
 MiNiFi can run as a Windows service. To do so, you must modify the `conf/bootstrap.conf` to set absolute paths for some properties. The properties are:
 
-* `lib.dir`
-* `conf.dir`
-* `nifi.minifi.config`
+* `lib.dir`  (e.g. `C:\\minifi-x.x.x\\lib`)
+* `conf.dir` (e.g. `C:\\minifi-x.x.x\\conf`)
+* `nifi.minifi.config` (e.g. `C:\\minifi-x.x.x\\conf\\config.yml`)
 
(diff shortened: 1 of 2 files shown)


jzonthemtn:
Would it be better to just use the 64bit version instead and change your documentation to be for 32bit? I'm guessing users might want 64bit version more often anyway.

ijokarumawak:
@jzonthemtn Hmm, I'm not sure, because the PR has been confirmed by at least two people and both didn't encounter issue.

I've tested it again. Even if on 64bit machine, if JDK (or JRE) for 32bit is installed the 32bit minifi.exe works. It didn't work before because I had only 64bit JDK.

ijokarumawak:
@jzonthemtn looking at more articles on current situation of 32 bit and 64 bit share, I'd prefer to go with 64 bit as default. Added another commit to make 64 bit as default. Does it look reasonable?

apiri:
yeah, 64 bit makes more sense at this stage.  As @ijokarumawak pointed out on the JIRA ticket, it seems that if a user has a 32-bit JRE installed, they would not experience issues.  The machine in particular that I was doing verification on seemed to have both 32 and 64 bit (presumably from some other application installed on the machine).  Anyway, this makes more sense to me.

apiri:
@ijokarumawak This was merged in 03aceefce523d3b957ee18c5fb1a595241dddc4d but I forgot to add the closing line.  Would you please be able to close this issue when you have the opportunity?

PR #114, opened 2018-02-02
MINIFI-434 - PullHttpChangeIngestor should preserve security properties

pvillard31:

### For all changes:
     in the commit message?

### For code changes:

### Note:

### DIFF
--- a/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/configuration/ingestors/AbstractPullChangeIngestor.java
+++ b/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/configuration/ingestors/AbstractPullChangeIngestor.java
@@ -27,11 +27,9 @@ import java.util.Properties;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
-
-
+import java.util.concurrent.atomic.AtomicReference;
 
 public abstract class AbstractPullChangeIngestor implements Runnable, ChangeIngestor {
-
 
     // 5 minute default pulling period
     protected static final String DEFAULT_POLLING_PERIOD = "300000";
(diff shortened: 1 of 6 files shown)


pvillard31:
This should probably be extended to more than just the security properties but this alone unlocks a situation when trying to perform secured S2S between agents and a NiFi instance. Note I'm not sure that's the best way to address it, happy to hear any feedback.

jzonthemtn:
This changes the current behavior, right?

pvillard31:
Hi @jzonthemtn - that's correct, if we want to preserve current behavior, we should set the property to true by default. But to be honest, I believe that, without this change, current mechanism prevent any use of secured communication. And if there is no security properties, then whatever the value is, the behavior is not changed. Happy to change the default value if desired.

jzonthemtn:
@pvillard31 I totally agree that this change is needed. I was even wondering if it should just be the standard behavior without an option. But I guess someone somewhere for whatever reason might not want it so an option is good. I was just thinking about how to best put it in the release notes for the next version that the default behavior will have changed for anyone currently with a manual process to revert the security settings. I'm not aware of anyone who has asked on the list (but I could have missed some) so maybe it won't affect many users.

pvillard31:
Agree with you. Not sure an option is needed at all.

apiri:
This looks good and functionality seems okay.  Would you mind adding a test that covers the functionality you are introducing with the override?  Just want to make sure it is something we maintain  and consider moving forward.

pvillard31:
@apiri I just pushed a commit adding a unit test for it. What's your opinion about offering this feature through an option? Wondering if the option makes sense... would someone want to override security properties with empty conf?

apiri:
@pvillard31 I think the option is okay.  I do not see this being the case for an empty conf but I could see it for the scenario where we are introducing a new set of security properties such as the case when a cert expires.  Seem like a worthwhile case?

pvillard31:
Yeah the option is OK. I think the discussion (for the future) is more about how to control parts of the configuration that are not related to the flow itself. With this ingestor and if the C2 instance is configured to pull the flow from NiFi templates (and later the NiFi Registry), what we'll get will only contain workflow information. I guess we could think of two pulling mechanisms: one for flow definition, one for core configuration of the agent. Or the C2 could be configured to merge the flow retrieved from templates/registry with core informations when requested by the agent... And the core information could be defined by "groups" of agents and this group membership would be sent by the agent when requesting the C2... ex: I'm belonging to the "windows-agent" group, give me the last version of the flow, and give me the core information for my group... Anyway :) having said that, this PR is probably fine as-is for now, and we have plenty of follow-up JIRAs.

apiri:
@pvillard31 Yep, fair concerns, and ultimately, I think to best align ourselves with making use of registry, our the YAML needs to only be the flow information and nothing else.  I think the thoughts you present seem inline with some of the initial ideas that have been tossed out there on the mailing lists or the wiki itself.  Anyway, thanks for this and we can start looking at additional JIRAs to start fleshing this out a bit more.

PR #115, opened 2018-02-05
MINIFI-433 - PG Variable Registry support in templates

pvillard31:

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced 
     in the commit message?

- [ ] Does your PR title start with MINIFI-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [ ] Is your initial contribution a single, squashed commit?

### For code changes:
- [ ] Have you ensured that the full suite of tests is executed via mvn -Pcontrib-check clean install at the root nifi-minifi folder?

### Note:

### DIFF
--- a/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
+++ b/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/ConfigTransformer.java
@@ -383,6 +383,7 @@ public final class ConfigTransformer {
             addTextElement(element, "name", processGroupSchema.getName());
             addPosition(element);
             addTextElement(element, "comment", processGroupSchema.getComment());
+            addVariables(doc, element, processGroupSchema.getVariables());
 
             for (ProcessorSchema processorConfig : processGroupSchema.getProcessors()) {
                 addProcessor(element, processorConfig);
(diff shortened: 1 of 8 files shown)


pvillard31:
**Do not merge as-is.**

I added unit tests, and it seems to behave as expected but when testing with a local build, variables are not included in the flow.xml.gz generated from the config.yml. I guess I missed something somewhere, but didn't find it so far. I still submitted the PR in case someone wants to have a look.

JPercivall:
Hey @pvillard31, took a look and I believe the problems you're seeing are due to variables at the root level not being accounted for in the template transformation. In ConfigSchemaFunction, for the root group level you're passing null for the variables[1]. So the check for "variables" in the recursive function below that will not have the variables for the root group.

The transformation works for the actual config transformation (from config.yml to flow.xml) because there's no special passing of the variables[2]. 

The reason the unit tests don't catch it because the toolkit unit test doesn't include a variable at the root level[3] (the ConfigMainTest though).

[1] https://github.com/apache/nifi-minifi/pull/115/files#diff-086bfd1ce3301edce34fd8af627c1470R88
[2] https://github.com/apache/nifi-minifi/pull/115/files#diff-a022622bcf128ab992d247f2a8694a0eR429
[3] https://github.com/apache/nifi-minifi/pull/115/files#diff-f9d5c0852f4099119785f1880d047d24R12282
[4] https://github.com/apache/nifi-minifi/pull/115/files#diff-76cab7aede126aa1ab9da5d56a4b5580R282

PR #117, opened 2018-02-28
MINIFI-424 Adding the ability to evaluate bootstrap properties in the…

JPercivall:
… config YAML prior to translation into flow.xml and nifi.properties

### For all changes:
- [X] Is there a JIRA ticket associated with this PR? Is it referenced 
     in the commit message?

- [X] Does your PR title start with MINIFI-XXXX where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [X] Has your PR been rebased against the latest commit within the target branch (typically master)?

- [X] Is your initial contribution a single, squashed commit?

### For code changes:
- [X] Have you ensured that the full suite of tests is executed via mvn -Pcontrib-check clean install at the root nifi-minifi folder?
- [X] Have you written or updated unit tests to verify your changes?
- [X] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? 
- [X] If applicable, have you updated the LICENSE file, including the main LICENSE file under minifi-assembly?
- [X] If applicable, have you updated the NOTICE file, including the main NOTICE file found under minifi-assembly?

- [X] Have you ensured that format looks appropriate for the output in which it is rendered?

### Note:

### DIFF
--- a/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
+++ b/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
@@ -1198,7 +1198,7 @@ public class RunMiNiFi implements QueryableStatusAggregator, ConfigurationFileHo
         final Properties transformationConfigProperties = getConfigTransformationProperties(bootstrapProperties);
 
         for(String name: transformationConfigProperties.stringPropertyNames()) {
-            defaultLogger.error("config property: name:[" + name + "] value:[" + transformationConfigProperties.getProperty(name) + "] ");
+            defaultLogger.info("config property: name:[" + name + "] value:[" + transformationConfigProperties.getProperty(name) + "] ");
         }
 
         try (InputStream inputStream = new FileInputStream(configFile)) {
--- a/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
+++ b/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
@@ -22,7 +22,7 @@ import org.apache.nifi.minifi.bootstrap.configuration.ConfigurationChangeCoordin
 import org.apache.nifi.minifi.bootstrap.configuration.ConfigurationChangeException;
 import org.apache.nifi.minifi.bootstrap.configuration.ConfigurationChangeListener;
 import org.apache.nifi.minifi.bootstrap.status.PeriodicStatusReporter;
-import org.apache.nifi.minifi.bootstrap.util.ConfigTransformer;
+import org.apache.nifi.minifi.bootstrap.configuration.transformation.ConfigTransformer;
 import org.apache.nifi.minifi.commons.status.FlowStatusReport;
 import org.apache.nifi.util.Tuple;
 import org.apache.nifi.util.file.FileUtils;
@@ -79,6 +79,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
+import static org.apache.nifi.minifi.commons.schema.common.CommonPropertyKeys.BOOTSTRAP_CONFIG_PROPERTY_PREFIX;
 
 /**
  * <p>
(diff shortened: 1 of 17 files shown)


minifi-commons/minifi-commons-schema/src/test/java/org/apache/nifi/minifi/commons/schema/serialization/SchemaLoaderTest.java
```diff
@@ -84,13 +85,92 @@ public void testUnsupportedVersion() throws IOException, SchemaLoaderException {
         }
     }
 
+    @Test
+    public void testMinimalConfigV3VersionUnusedProperties() throws IOException, SchemaLoaderException {
+        Properties inputProperties = new Properties();
```
apiri: Minor, but could we also have these tests drawing from a sample bootstrap.conf in lieu of just seeding the properties.  With some of the work that is underway surrounding C2 would not want to miss coverage of this functionality.

minifi-toolkit/minifi-toolkit-configuration/src/main/java/org/apache/nifi/minifi/toolkit/configuration/ConfigMain.java
```diff
@@ -106,10 +108,22 @@ public static void printValidateUsage() {
     }
 
     public int validate(String[] args) {
-        if (args.length != 2) {
+        if (args.length != 2 && args.length != 3) {
             printValidateUsage();
```
jzonthemtn: Does `printValidateUsage()` need updated?

minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/RunMiNiFi.java
```diff
@@ -1176,11 +1191,18 @@ public boolean accept(final File dir, final String filename) {
...
     @SuppressWarnings({"rawtypes", "unchecked"})
     public void start() throws IOException, InterruptedException {
+        Properties bootstrapProperties = getBootstrapProperties();
+
+        final String confDir = bootstrapProperties.getProperty(CONF_DIR_KEY);
+        final File configFile = new File(bootstrapProperties.getProperty(MINIFI_CONFIG_FILE_KEY));
+        final Properties transformationConfigProperties = getConfigTransformationProperties(bootstrapProperties);
+
+        for(String name: transformationConfigProperties.stringPropertyNames()) {
+            defaultLogger.error("config property: name:[" + name + "] value:[" + transformationConfigProperties.getProperty(name) + "] ");
```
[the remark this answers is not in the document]
JPercivall: Most definitely better as `info`

JPercivall:
@apiri @jzonthemtn sorry for not responding in a more timely fashion. I've addressed your code comments and rebased with the most recent commit. Thank you for the review!

@apiri as for rolling this into EL as Variable Registry interface, I don't think that is appropriate as this covers things that are outside the scope of evaluated expressions. For example, this allows the user to set elements that are normally in the `nifi.properties`, or not exposed to EL (such as a S2S url and port).

JPercivall:
Hey @apiri, just circling back if there's anything more you want me to do/clarify.

apiri:
@JPercivall whoops, sorry this got lost in the mix.  Refreshing myself I appear to have been a bit unclear in my original messaging.  I would say I am not advocating for the inclusion of EL, however the VariableRegistry still seems like the right abstraction.  For similar works, consider something like Notifications configured via bootstrap in NiFi https://github.com/apache/nifi/blob/master/nifi-bootstrap/src/main/java/org/apache/nifi/bootstrap/NotificationServiceManager.java#L74.

Given that we are managing the transformation process, we can inspect and manipulate config files as needed.  In the minifi case, we get to inject the appropriate bits where needed and could make use of the framework functionality already in-place.

Longer term, looking at the issues/proposals out there, I have concerns that this would be yet another config process kind of baked into the code that might make it complicated to extract out to another location when configs are generated/delivered via other means.

JPercivall:
Re-reading @apiri's input and seeing more of the activity with the other projects I don't think this is the right way to go at this time. I still think we need a better way to oversee what makes up the "nifi.properties" portion and provide better runtime configuration of the instance but doing so out of band with MiNiFi C++ and NiFi would not be the best approach. 

Closing

PR #120, opened 2018-03-28
MINIFI-448 C2 server functionality and internal interfaces

kevdoran:
The C2 service layer is the part of the C2 framework that backs the REST API to provide the business logic.

This commit also defines a few preliminary provider interfaces for persistence based on what is required by the service layer. A reference, in-memory implementation of the provider interfaces is included.

--- 

### For all changes:
     in the commit message?

### For code changes:
- [x] Have you written or updated unit tests to verify your changes?

### Note:


minifi-c2/minifi-c2-framework/src/test/groovy/org/apache/nifi/minifi/c2/core/service/StandardC2ProtocolServiceSpec.groovy
```diff
@@ -0,0 +1,164 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.minifi.c2.core.service
+
+import org.apache.nifi.minifi.c2.api.provider.heartbeat.HeartbeatPersistenceProvider
+import org.apache.nifi.minifi.c2.model.*
+import spock.lang.Specification
+
+class StandardC2ProtocolServiceSpec extends Specification {
```
jzonthemtn: Just curious if there was a technical reason for using Groovy for the test? I don't know a whole lot about groovy so I could be missing something that's otherwise obvious.

minifi-c2/minifi-c2-framework/src/main/java/org/apache/nifi/minifi/c2/api/provider/device/DevicePersistenceProvider.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * NOTE: Although this interface is intended to be an extension point, it is not yet considered stable and thus may
+ * change across releases until the the C2 Server APIs mature.
+ *
+ * TODO, we may want to consider creating a separate entity model rather than reusing the REST API object model.
+ * Currently, this design assumes the Provider implementation will do that translation.
+ * This requires adding a dependency on minifi-c2-commons here for the data model.
+ */
+public interface DevicePersistenceProvider extends Provider {
```
bbende: Would it make sense to have a base interface kind of like the Spring Data repositories, something like:

```
interface PersistenceProvider<T> {
  long getCount();
  T save(T t);
  List<T> getAll();
  Optional<T> get(String id);
  void delete(T t);
}
```

Then some of the interfaces just become empty:
```
interface DevicePersistenceProvider extends PersistenceProvider<Device> {
    // add any specific methods that only make sense for devices
}
```

The only case where it doesn't fully work is the AgentPersistenceProvider which has the CRUD operations for multiple domain objects in a single interface, which makes sense to enforce that they are persisted to the same provider. 

Maybe that interface could provider sub-services for each domain type?
```
interface AgentPersistenceProvider {
   PersistenceProvider<AgentClass> getAgentClassPersistenceProvider();
   PersistenceProvider<AgentManifest> getAgentManifestPersistenceProvider();
   PersistenceProvider<Agent> getAgentPersistenceProvider();
}
```
So the only pluggable provider is the top-level AgentPersistenceProvider, but then it is up to that provider to provider implementations of the sub-providers.

Feel free to tell me this makes no sense, just throwing out ideas.

minifi-c2/minifi-c2-framework/src/main/java/org/apache/nifi/minifi/c2/core/service/StandardC2Service.java
```diff
@@ -0,0 +1,494 @@
...
+public class StandardC2Service implements C2Service {
+
+    private static final Logger logger = LoggerFactory.getLogger(StandardC2Service.class);
+
+    private final AgentPersistenceProvider agentPersistenceProvider;
+    private final DevicePersistenceProvider devicePersistenceProvider;
+    private final OperationPersistenceProvider operationPersistenceProvider;
+    private final Validator validator;
+
+    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
```
bbende: Do we think the read/write lock approach is too limiting for C2?

I know we did this in registry to ensure consistency when interacting with more than one data store, but the trade-off is that it reduces the concurrency a lot because all threads are blocked while one thread has the write lock. 

I don't know the level of concurrent requests we are anticipating here so it may be a non-issue, but just wanted to throw it out there. This also may be totally fine for initial implementation, and we can modify/improve later.

I also haven't read through all the methods below yet, but if there isn't a case where a single method performs operations across multiple persistence providers, then we could probably make the locks more granular and have a lock for each persistence provider so that agent calls are not blocked by device calls, etc.

minifi-c2/pom.xml
```diff
@@ -37,10 +37,30 @@ limitations under the License.
...
+                            <goal>testCompile</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
     <properties>
+        <nifi.registry.version>0.1.0</nifi.registry.version>
```
bbende: Same property below

minifi-c2/minifi-c2-framework/src/main/java/org/apache/nifi/minifi/c2/api/provider/device/DevicePersistenceProvider.java
kevdoran: > Would it make sense to have a base interface kind of like the Spring Data repositories

I think this makes a lot of sense. I'm going to create this base interface type as you suggested. Are there any other methods you can think of that should be included the base interface? (It has everything needed for the current framework invocations, but expecting third-parties might provide an implementation, is there anything else we can anticipate wanting provided?)

> So the only pluggable provider is the top-level AgentPersistenceProvider, but then it is up to that provider to provider implementations of the sub-providers.

Yeah I like this approach. Going to do that for now and we can refine it (if needed) as these provider interfaces mature a bit.

minifi-c2/minifi-c2-framework/src/main/java/org/apache/nifi/minifi/c2/core/service/StandardC2Service.java
kevdoran: Great points, @bbende. Thanks for the input.

This is probably too limiting as you point. One of the things I like about trying to handle concurrency in the service layer, as we did for registry, is that there is less burden placed on the implementations of the persistence providers that are called from the service methods (though if they are DB backed most DB engines will take care of that for you). But I agree, as implemented here, it's too much of a performance trade off if we want to support high concurrency / number of clients for a C2 server instance.

I'll take another pass at this and look at it on more of a case-by-case basis.

kevdoran:
Thanks for tying review, @apiri. I’m going to take some time to implement some of @bbende’s suggestions prior to merging and ten will update this PR.

minifi-c2/minifi-c2-framework/src/test/groovy/org/apache/nifi/minifi/c2/core/service/StandardC2ProtocolServiceSpec.groovy
kevdoran: Good question, @jzonthemtn. There is no technical/functional reason these tests couldn't be written as Java junit tests. The groovy framework I'm using here, [Spock](https://github.com/spockframework/spock), offers a few advantages:

- Detailed error messages for failed tests/assertions
- A reasonable terse/structured syntax that allows tests to serve as feature/method requirements/specifications. This was really beneficial in writing [StandardC2ServiceSpec](https://github.com/apache/nifi-minifi/pull/120/files#diff-f485f870b41384487a26ff065fd25467), which is extremely repetitive. I wanted the tests to be short and self-descriptive so I could keep track of what combinations were covered.
- Powerful mocking capabilities (with some advantages over Java libs such as Mockito)
- built-in test templating and data driven parameterization (although I did not use that feature in this test cases).

jzonthemtn: Thanks for the details! Sounds like something worth checking out. I remember seeing a discussion not too long ago on the mailing list concerning JVM-based languages in the code base. (I don't remember if anything actually came out of it or not.) Anyway, I think it's pretty cool because otherwise I probably wouldn't have been exposed to this.

kevdoran:
@bbende I made some changes to concurrency locking in the service layer and the persistence APIs based on your feedback. For now, I'm not enforcing that the same provider is used for all the agent* persistence providers as those are not really extension points yet. Prior to making that an extension API, we should decide the granularity of those persistence providers (all or nothing, or individual, and if individual, which ones are forced to be grouped). 

If you have a chance to re-review, that would be great, thanks!

bbende:
Looks good to me, I'm a +1... nice work!

kevdoran:
Thanks! will get this merged....

kevdoran:
merged

PR #121, opened 2018-03-29
MINIFI-449 Support for converting VersionedFlowSnapshot to YAML via m…

bbende:
…inifi-toolkit

- Upgrading Jackson databind to 2.9.1 for consistency with other Jackson modules

For testing this I did the following...

- Made a GetFile -> RPG flow inside of a process group (same as you would normally make a template of)
- Started version control and saved this PG to my NiFi Registry instance
- Used NiFi CLI to export the JSON (from nifi-toolkit)
- Ran the new transform-vfs command using the above JSON as the input

The NiFi CLI commands would be something like the following:
```
registry list-buckets -u <registry-base-url>
registry list-flows -u <registry-base-url> -b <bucket id>
registry export-flow-version -u <registry-base-url> -f <flow-id> -o my-flow.json
```

### DIFF
--- a/minifi-toolkit/minifi-toolkit-configuration/pom.xml
+++ b/minifi-toolkit/minifi-toolkit-configuration/pom.xml
@@ -36,4 +36,18 @@ limitations under the License.
             <artifactId>nifi-framework-core</artifactId>
         </dependency>
     </dependencies>
+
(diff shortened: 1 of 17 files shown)


apiri:
reviewing

PR #158, opened 2019-07-09
MINIFI-478 Fix broken OrderedProperties and OrderedPropertiesTest

shonzilla:
This fix is not specific to a platform or Java version.

This test was previously passing by plain coincidence. Here's a more detailed explanation:
- The only ordered data structure in `OrderedProperties` was `linkedHashSet` happened to be updated but never read and therefore it was never consulted for reading properties in ordered manner, i.e. in the same order they were put/set.
- The expectation and the contract of `OrderedProperties` is that overridden `store(OutputStream out, String comments)` method stores properties in ordered manner. However, this doesn't happen without this PR.
- The order in which properties are stored does not depend on `keys()` method but `entrySet()` method as it can be witnessed along the following call path `OrderedProperties.store` -> `Properties.store` -> `Properties.store0` where the entries `(Ordered)Properties` are read via `entrySet()`. Without this PR the default `Properties entrySet()` implementation is used to read properties.
- The order in which properties are read in `Properties entrySet()` which is based on underelying `ConcurrentHashMap.entrySet()` where entries are **unordered**.
- The reason why `OrderedPropertiesTest` was ever passing (on earlier Java versions) is the combination of a small test data set - only three properties with specific property names and the hash function implementation within the underlying `ConcurrentHashMap.putVal`, that stored those property names in such a way that the corresponding `ConcurrentHashMap.entrySet()` returned them in order by coincidence. The fact that `OrderedProperties.sets()` and any order wasn't explicitely used was not realized by neither implementor nor anyone else by this same coincidence.
- Due to limitations of my current computer setup, I didn't bother to compare side-by-side the hash function implementations of `ConcurrentHashMap.putVal` across Java versions so this is left as an exercize to the reader. However, note that hash function implementations may change from Java version to version since their role is to provide statistically normalized data (i.e. hash key) distribution across a hash map without having to guarantee the same mapping from name to a hash map slot.

### For all changes:
     in the commit message?

### For code changes:
- [x] Have you written or updated unit tests to verify your changes?
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](https://www.apache.org/legal/resolved.html#category-a)? 

### Note:

### DIFF
--- a/minifi-bootstrap/src/test/java/org/apache/nifi/minifi/bootstrap/util/OrderedPropertiesTest.java
+++ b/minifi-bootstrap/src/test/java/org/apache/nifi/minifi/bootstrap/util/OrderedPropertiesTest.java
@@ -36,20 +36,26 @@ public class OrderedPropertiesTest {
         orderedProperties.setProperty("prop1", "origVal1");
         orderedProperties.setProperty("prop2", "val2", "#this is property 2");
         orderedProperties.setProperty("prop3", "val3");
+        orderedProperties.setProperty("prop4", "val4");
+        orderedProperties.setProperty("prop5", "val5", "#this is property 5");
+        orderedProperties.setProperty("prop3", "newVal3");
         orderedProperties.setProperty("prop1", "newVal1");
         ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
         orderedProperties.store(byteArrayOutputStream, PROPERTIES_FILE_APACHE_2_0_LICENSE);
 
-        try (BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(new ByteArrayInputStream(byteArrayOutputStream.toByteArray())));
+        try (BufferedReader actualReader = new BufferedReader(new InputStreamReader(new ByteArrayInputStream(byteArrayOutputStream.toByteArray())));
              BufferedReader expectedReader = new BufferedReader(new InputStreamReader(OrderedPropertiesTest.class.getClassLoader().getResourceAsStream("orderedPropertiesExpected.properties")))) {
             String expectedLine;
             while((expectedLine = expectedReader.readLine()) != null) {
-                String actualLine = bufferedReader.readLine();
+                String actualLine = actualReader.readLine();
+                // assertEquals(expectedLine, actualLine);
                 if (!"#Tue Feb 21 11:03:08 EST 2017".equals(expectedLine)) {
                     assertEquals(expectedLine, actualLine);
+                } else {
+                    System.out.println("Skipping timestamp comment line");
                 }
             }
-            assertNull(bufferedReader.readLine());
+            assertNull(actualReader.readLine());
         }
     }
 }
--- a/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/OrderedProperties.java
+++ b/minifi-bootstrap/src/main/java/org/apache/nifi/minifi/bootstrap/util/OrderedProperties.java
@@ -25,27 +25,36 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
+import java.util.AbstractMap;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 public class OrderedProperties extends Properties {
     private final Map<String, String> textBeforeMap = new HashMap<>();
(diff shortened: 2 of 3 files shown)


apiri:
Hi @shonzilla 
Code looks fine to support both 8 and 11.  The implementation of the store0 method does actually change from 8->11 as seen in the the relevant sources:

8: http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/Properties.java#l832
11: https://github.com/unofficial-openjdk/openjdk/blob/jdk/jdk/src/java.base/share/classes/java/util/Properties.java#L926

It is my belief leaving keys() as is and then introducing the overridden entryset as provided by you should cover both.  Does that thought process seem reasonable to you as well?

minifi-bootstrap/src/test/java/org/apache/nifi/minifi/bootstrap/util/OrderedPropertiesTest.java
```diff
@@ -36,20 +36,26 @@ public void testOrderedProperties() throws IOException {
...
         orderedProperties.store(byteArrayOutputStream, PROPERTIES_FILE_APACHE_2_0_LICENSE);
 
-        try (BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(new ByteArrayInputStream(byteArrayOutputStream.toByteArray())));
+        try (BufferedReader actualReader = new BufferedReader(new InputStreamReader(new ByteArrayInputStream(byteArrayOutputStream.toByteArray())));
              BufferedReader expectedReader = new BufferedReader(new InputStreamReader(OrderedPropertiesTest.class.getClassLoader().getResourceAsStream("orderedPropertiesExpected.properties")))) {
             String expectedLine;
             while((expectedLine = expectedReader.readLine()) != null) {
-                String actualLine = bufferedReader.readLine();
+                String actualLine = actualReader.readLine();
+                // assertEquals(expectedLine, actualLine);
```
apiri: should remove this errant comment

apiri:
Hi @shonzilla,

Just wanted to see if you had any thoughts on the comments I left so we can get this wrapped up.  Thanks!

apiri:
Going to make the suggested changes and include as I believe the functionality to be correct as discussed.

shonzilla:
Hi @apiri, sorry for not getting back in timely fashion. After some frantic time at work and subsequent vacation (mostly offline, no-laptop), I got back into the pickle with no time for open-source. :wink: 

Thanks for your patience and for removing the superfluous comment for me with your amended merge commit.

I've re-checked the JDK source code changes between versions 8 and 11 and indeed - `Properties.store` (i.e. the underelying private `store0` method) has migrated from iterating over its `keys` to iterating over its `entrySet`. This storing of key-value pairs with incl. comments to an output stream is part of the _current_ unit test before it loads already ordered properties from the said output stream. If I remember correctly, this test coverage is sufficient for the little use of `OrderedProperties` in the MiNiFi project.

That's why one needs both to keep both the original overridden `keys` and the newly overridden `entrySet` method, contrary to what I originally thought. However, one should keep in mind that in both cases the requirement is that either keys (in JDK 8) or entries (key-value pairs, used in JDK 11) must be returned in ordered manner, i.e. in the same order that the individual properties/entries were put into an `OrderedProperties` instance. Without additional testing (on JDK 8) I cannot verify and hence cannot guarantee that the `keys` method returns keys in order that they were put into `OrderedProperties`.

I'm temporarily down to Rasberry Pi 3+ with just 1GB (incl. OpenJDK 11 though) so it may be problematic to verify this by building NiFi... and it would certainly hurt without closing the web browser first. :laughing:

Speaking of Rasberry Pi - it might be an interesting experiment though to get MiNiFi or NiFi (to build and) run on it. One could build for example a self-hosted and highly flexible service comparable to IFTTT but isolated from the public internet, used to make one's home automation smarter etc.
Are you aware of anyone doing this with MiNiFi or NiFi?

A quick web search revealed that (obviously) I'm not the first man to think about running NiFi on a RasPi and it's great to already know that it's possible and maybe even feasible. :nerd_face:

apiri:
@shonzilla No worries.  These things happen, thanks for following up and your contributions.  I think we are on the same page and thus why I moved forward with the merge (and cleaned up the errant comment).

We've had several folks that have done some neat projects and exercises centered around Pis.  I would encourage you to scope out our lists and optionally post to reach out to the broader community (http://nifi.apache.org/mailing_lists.html).  Builds that support ARM are even enabled via our bootstrap process on our C++ impl (https://github.com/apache/nifi-minifi-cpp/).  A unique twist on integration with other services that comes to mind is http://www.opensourcedad.com/apache/minifi-cpp/2016/12/18/poop-scale.html

PR #196, opened 2021-01-24
[MINIFI-546] Upgrade to Nifi 1.12 and Java11 via jaxb dependencies

jomach:
I all, 

I created two PR's for this but I could not complete the build.
Can someone help on:
* StatusConfigReporterTest -> ignored tests as I don't know If is due to the api change
* AbstractTestSecure -> are failing and I could not find the certificates on how to load, would be great if someone could help. 
* I have another branch which relies on XML Jackson instead of JAXBContext.newInstance. Let me know if I should create a pr with this code instead.


gruselglatz:
I think the Nifi Team is integrating Minifi into the Nifi Codebase, to bring it on the same version level.
See: https://issues.apache.org/jira/browse/MINIFI-422

